Issue #173, opened 2019-01-14
Reference: hide internal methods


distantnative:
We have a few methods that are really only intended for internal use. We should have a DocBlock flag to not include them in the auto generated reference.

distantnative:
Use `@internal`

Issue #137, opened 2018-12-18
[Docs] Ignore deprecated methods in cheatsheet


bastianallgeier:

distantnative:
Related: https://github.com/k-next/getkirby.com/issues/173

PR #186, opened 2019-01-15
Hide deprecated and internal methods in reference

distantnative:
A first draft of how we could filter those methods using `@internal` and `@deprecated`.

Solves #137 
Solves #173


bastianallgeier:
Yay!

PR #241, opened 2019-01-22
Fixes #240

### LINKED ISSUE #240: Guide: Documentation for user blueprints not correct
https://forum.getkirby.com/t/user-can-change-his-own-role-in-panel-when-changerole-is-turned-off/12341

seanbuchholz:
Update permissions to proper syntax. Key `users` (plural) is used under `access` key, but the corresponding top-level permission is keyed as `user`, singular.


texnixe:
Thanks for fixing 👍

PR #245, opened 2019-01-23
sections extension: wrong context for this

mullema:
In the examples "this" is in the wrong context and therefore there is no output (and no error message). I suggest to use arrow functions instead.


distantnative:
Very true. But just a thought: Are arrow functions supported by all browsers without running babel over them?

mullema:
not really. But as far as I can see are arrow functions supported from the same browsers that support `fetch()`. 
https://caniuse.com/#feat=arrow-functions
https://caniuse.com/#feat=fetch
and `fetch()` seems to be a requirement for the panel?

bastianallgeier:
LGTM!

PR #282, opened 2019-02-02
Add user field migration

distantnative:
@texine Pleas have a look! Thanks :)


distantnative:
@texnixe ping

distantnative:
Mainly cause I am afraid someone will run this script before really updating all blueprints. And with this way it shouldn't be a problem then in which order it happens.

texnixe:
Makes sense. I close my case.

PR #324, opened 2019-02-13
Fix links to $field->yaml() and $field->toStructure()

Splendorr:
These are under docs/reference/templates/field-methods, not docs/reference/panel/field-methods. I _think_ I got the folder tree links right, because they need to go back down two levels? You should double-check me, though!


texnixe:
Thank you!

PR #393, opened 2019-03-10
Docs for filtering by date

frederikbrudy:
Date field doesn’t return a Unix timestamp anymore. Update docs to reflect how to filter by date.


frederikbrudy:
I also just noticed the ["date field is special" remark](https://github.com/getkirby/getkirby.com/blob/7f5243dc68922002cc9adf5c0153cf1098ba85ad/content/1_docs/2_cookbook/0_search/0_filtering/recipe.txt#L184) in the docs. I haven't fully tested it, but my understanding is that this is not the case anymore. Maybe this needs to be fixed as well.

texnixe:
Thanks a lot for fixing this!

PR #414, opened 2019-03-12
Refactor and improve API docs

distantnative:


content/1_docs/1_guide/8_users/3_permissions/guide.txt
```diff
@@ -52,14 +57,16 @@ permissions:
...
 ```
-You don't have to set all of those permissions for each role though. Each permission is set to true by default, if not specified otherwise. You can also enable or disable entire sections by passing `true` or `false` instead of an array of definitions for each action. A more simple permission setup for an editor could be something like this:
+
+<info>
+The difference between `users` and `user` permissions is that the later only applies to the user itself, while the former grants these permissions over all users.
+</info>
+
+You don't have to set all of those permissions for each role though. Each permission is set to `true` by default, if not specified otherwise.
+
+You can also enable or disable entire sections by passing `true` or `false` instead of an array of definitions for each action. A more simple permission setup for an editor could be something like this:
```
texnixe: For me that example is not really clear to explain what enabling or disabling entire sections is supposed to mean.

content/1_docs/1_guide/8_users/3_permissions/guide.txt
```diff
@@ -52,14 +57,16 @@ permissions:
     changeRole: false
     delete: false
     update: false
-  files:
-    create: true
-    changeName: true
-    delete: true
-    replace: true
-    update: false
+  users: false
```
texnixe: The users options are missing if these are supposed to be all possible actions.

distantnative: ✅

PR #448, opened 2019-03-28
Feature/cleanup bugfix etc

fabianmichael:
Most important breaking change: Custom development host is now stored in `config.json` instead of `config.js`.

Apart from that, we now got content-based cache-busting instead of modified time, which will make caching way more effective.

The assets/dev/ folder is gone.

All dependencies have been updated to the lastest version.

We got linting and backstop testing.

Some visual bug fixes.


distantnative:
LGTM – but needs some conflicts to be resolved first.

PR #473, opened 2019-04-07
Add database reference

texnixe:


bastianallgeier:
I think we need to work on this manually as well. The DB reference isn't really helpful because most of the features are achieved with magic calls. The query docs are misleading as well because you normally do something like this: 

```php
$db = new Database([...]);

$db->table('users')->select('*')->where('created', '>', '2012-12-12')->all()
```

PR #490, opened 2019-04-14
Correct grammar in Escape class docs

simlawr:
The use of “the” with “Zend Escaper” isn’t grammatically correct.


bastianallgeier:
Thanks for your help 🙌

PR #501, opened 2019-04-16
Restructure content guide

texnixe:


content/1_docs/1_guide/4_content/2_creating-pages/guide.txt
```diff
@@ -54,29 +47,28 @@ Each page folder is accessible via its URL, e.g. you can access the projects fol
 
 URLs can be modified using (link: docs/guide/routing text: Kirby's router).
 
-## Creating pages in the Panel
+## Creating drafts manually
```
lukasbestle, line 64: Maybe we should reorder the sections to have the information on the Panel first and the manual and programmatic variants second (for advanced users). I think many users wouldn't want to know the internals, they'd rather get started with the Panel.

Besides that, the changes look great!

texnixe: Yes, makes sense

lukasbestle (approved):

PR #507, opened 2019-04-18
Fix users field type

ghost:
It's `users` not `user`  
https://getkirby.com/docs/reference/panel/fields/users


distantnative:
Thanks for the help!

PR #517, opened 2019-04-19
Fix example code in Validators reference

simlawr:
Uppercase all class names to match Kirby 3 naming scheme.


texnixe:
Thanks a lot!

PR #519, opened 2019-04-19
Fix example code in V::size() reference

simlawr:
Another class name that should be uppercase.


simlawr:
#520 has a more comprehensive set of classes.

PR #527, opened 2019-04-25
Add changeRole permission to documentation

Jayshua:
Not sure if this is intentionally undocumented?


bastianallgeier:
No, it's not intentional. Thank you for spotting and fixing it!

PR #539, opened 2019-05-02
Language Switch: add hreflang attribute

mrflix:
`hreflang` indicates the language the linked page is in: https://support.google.com/webmasters/answer/189077?hl=en

There are a lot of ways to indicate the target page's language. I find this the simplest one. I think its a good idea to add it to the examples. If you don't feel the same feel free to close this pull request.

I left it out in the third case because that one is a bit more complicated – the language of the eventually linked page is ambiguous.


bastianallgeier (approved):
Lovely!

PR #552, opened 2019-05-11
Changes for 3.1.2

distantnative:


content/1_docs/3_reference/7_plugins/1_extensions/0_api/extension.txt
```diff
@@ -0,0 +1,83 @@
...
+### Access to `$kirby`
+
+API endpoints can also be registered as callbacks. Within the callback they have full access to the current Kirby instance.
+
+
+```php "/site/plugins/your-plugin/index.php"
+Kirby::plugin('your/plugin', [
+  'api' => function ($kirby) {
+    return [
+      // ...
```
texnixe: Could we replace the dots with an example of what to add here?

content/1_docs/1_guide/14_cache/guide.txt
```diff
@@ -26,6 +26,7 @@ return [
 ```
 <info>
 The page cache files are stored in the `/site/cache` folder.
+Page requests will only be cached if there are params in the URL.
```
texnixe: **no** params...

distantnative:
@texnixe changes/fixes added

PR #568, opened 2019-05-24
Changed German PHP docs link to English link

PaulMorel:


texnixe:
Thanks for spotting and correcting this.

PR #586, opened 2019-06-13
Github link to edit a cookbook recipes

yoanmalie:


texnixe:
Yes, that makes sense, been wanting to add that anyway. Thanks.

PR #602, opened 2019-06-26
Add Plugin Workflow cookbook recipe

yandodov:
@texnixe please check if the recipe is OK.


content/1_docs/2_cookbook/9_setup/0_plugin-workflow/recipe.txt
```diff
@@ -0,0 +1,250 @@
...
+### PHPUnit
+
+If you don't have PHPUnit, follow the instructions on (link: https://phpunit.de/ text: their site) to install it or if you use XAMPP, check (link: https://medium.com/oblik-studio/how-to-set-up-phpunit-on-xampp-and-windows-dad92dc11335 text: this article).
+
+### Kirby
+
+We'll need a Kirby installation too. It can house all of our plugins and since each one will have its own content (although it can use the installation's), we'll use the (link: https://github.com/getkirby/plainkit text: plainkit template) which is very minimal:
+
+```bash
+cd C:/xampp\htdocs/
```
texnixe: This should be OS agnostic

texnixe:
@hdodov Greate recipe! I've added a patch PR with some wording style changes, let me know if it's ok. Don't know how I can make a  PR against a PR...

distantnative:
I would suggest a more specific title, plugin workflow sounds too broad to me - more like what we have at https://getkirby.com/docs/guide/plugins/

yandodov:
I tried to fix a spelling mistake but my fork's history and the origin's history have diverged. What do we do now? I'm kind of new to contributing and pull requests.

**Edit:** In [my fork's commit history](https://github.com/OblikStudio/getkirby.com/commits/master) there's both my initial commit with the recipe _and_ the same commit by @texnixe. On the origin repo, there's only the latter. I guess that's where the problem comes from. What do I have to do to fix this? Perhaps a force pull from the origin so the histories line up?

texnixe:
Probably, but I had problems myself with modifying your PR yesterday, I'm not sure what would be the best way. If a force pull doesn't work, I'd just fork the current state again.

texnixe:
I've fixed it now. Try a force pull into your fork.

PR #625, opened 2019-07-22
update email field article

theresa-g:
added missing possibilities to encode email addresses. taken from: https://k2.getkirby.com/docs/cheatsheet/panel-fields/email


texnixe:
Thanks for contributing. I corrected some style issues, like use of classnames etc.

PR #642, opened 2019-08-04
Add chapter about Kirby and privacy

texnixe:
Please check if this is complete. 

The chapter also needs a new icon.

### DIFF
--- a/content/1_docs/1_guide/20_kirby-and-privacy/guide.txt
+++ b/content/1_docs/1_guide/20_kirby-and-privacy/guide.txt
@@ -12,7 +12,7 @@ Text:
 <info>
 The following information might be incomplete. Your site may collect or process personal data in other ways.
 
-We cannot give any legal advice what this means for you and your sites. If you are unsure what you need to do to comply with GDPR in particular or any other national privacy laws in your country, please consult a lawyer.
+We cannot give any legal advice what the following information means for you and your sites. If you are unsure what you need to do to comply with GDPR in particular or any other national privacy laws in your country, please consult a lawyer.
 </info>
 
 ## Preface
@@ -37,7 +37,11 @@ The Panel sets a session cookie for every visitor who uses the Panel to keep tra
 
 ## Licensing
 
-To register your license, the Panel connects to our licensing server to verify your license. When you do so, the license server stores the domain for which the license is registered on our license server.
+To register your license, the Panel connects to our licensing server **once** to verify your license. When you register your license, the following information is transmitted to the server:
+
+- license key
+- email address
+- domain of the Kirby installation
 
 ## Brute-force protection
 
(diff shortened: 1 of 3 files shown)


content/1_docs/1_guide/20_kirby-and-privacy/guide.txt
```diff
@@ -0,0 +1,63 @@
...
+## Plugins and custom code
+Your sites may store or process additional personal data depending on the Kirby plugins and custom code you are using. For example, some plugins like contact form plugins also use sessions for technical reasons.
+
+## Contact forms
+Your sites also process personal data once a contact form is submitted, a blog comment gets stored or files get uploaded by visitors. The same also applies to similar custom site features. The data you store in your content files may also contain personal data.
+
+## Hosting providers
+Data might also get stored and processed by your hosting provider. What sort of data they store and process depends on your contract with the hosting provider.
+
+
```
lukasbestle, line 73: Maybe a horizontal line here to separate the following paragraph from the text (or an info box for that paragraph).

content/1_docs/1_guide/20_kirby-and-privacy/guide.txt
```diff
@@ -0,0 +1,63 @@
...
+The user credentials are transmitted via HTTP. Therefore, we strongly recommend to use TLS encryption for your sites to protect the passwords and other personal data of your visitors and users.
+
+## Panel
+
+The Panel sets a session cookie for every visitor who uses the Panel to keep track of the currently logged in user. Additionally, the Panel stores unsaved changes in `localStorage`.
+
+## Licensing
+
+To register your license, the Panel connects to our licensing server to verify your license. When you do so, the license server stores the domain for which the license is registered on our license server.
```
lukasbestle: This doesn't really fit into the list of Kirby features that process data. But to be honest I don't know where it would fit better.

We should also add which information in particular gets transmitted to the server:

- license key
- email address
- domain of the Kirby installation

and when (only once when the license is registered).

content/1_docs/1_guide/20_kirby-and-privacy/guide.txt
```diff
@@ -0,0 +1,63 @@
...
+
+
+----
+
+Text:
+
+<info>
+The following information might be incomplete. Your site may collect or process personal data in other ways.
+
+We cannot give any legal advice what this means for you and your sites. If you are unsure what you need to do to comply with GDPR in particular or any other national privacy laws in your country, please consult a lawyer.
```
lukasbestle: Maybe "what the following information means for you and your sites" so that the disclaimer applies to everything.

lukasbestle:
I have reorganized the sections a bit and added a few more details. I think it's ready to publish once the icon is ready. 👍

lukasbestle (approved):

texnixe:
@lukasbestle From a structural point of view, your reorganisation makes a lot of sense. Personally, I would have preferred to stick with the flat structure because of the TOC for easy reference.

lukasbestle:
Good point. 🤔

I personally don't think the TOC is necessary for that page (not that long anyway) – so it's more important IMHO to have a clear structure that's easy to understand. The text is more suited to read in one go anyway.

But I'm definitely open for a more flat structure in case you think it would be easier to read.

texnixe:
Yes, I know it is not a very long read. My thinking is more that this is the type of content you might come back to every now and then for quick reference ("What were all those things I have to keep in mind?"), and in that case the TOC would serve as a kind of checklist.

lukasbestle:
Why don't we add a checklist at the top then? :)

texnixe:
🤷‍♀️ I give up (and I knew you would say that).

lukasbestle:
Oh, I didn't want to push my opinion through. Feel free to make any changes you think are appropriate. :)

texnixe:
All good, I'm usually all for structuring stuff, but I really like these TOCs because they make scanning and accessing stuff easier.

In any case, we need an icon and @bastianallgeier should probably also still approve.

lukasbestle:
Maybe we could also make the TOC plugin configurable so it can be displayed even for heading structures like this one.

distantnative:
Smallest comment: didn’t we ban the ampersand?

texnixe:
Did we? Isn't that discrimination?

lukasbestle:
We actually have a lot of them in our styleguide. :P

distantnative:
I thought we said once not to use them, but I might be mistaken. I get reminded a lot at my main job by our senior editor to never use them.

PR #645, opened 2019-08-05
Fixed kirbytext links

trych:
Quick fix for some links in the migrate guide. Links should point to the `kirbytext` hooks, but were pointing to the `kirbytags` hooks.


texnixe:
Thanks! Guess that was a leftover because the kirbytext hooks were only added shortly before the release.

PR #647, opened 2019-08-06
Feature/some more fixes

fabianmichael:
This adds a few fixes for some styling issues and a JavaScript-issue, where code highlighting was not applied after content was loaded using AJAX. The problems might seem simple at first, but required rather extensive changes. Please test before merging.


texnixe:
Closing this as it looks like most of these changes have already been incorporated.

PR #656, opened 2019-08-12
Add sortBy blueprint examples to Pages field and Sorting Cookbook

Splendorr:
Hi! I spent a couple of hours searching the docs and forum before figuring out how to use sortBy as a blueprint option. Specifically, I wanted to sort a `type: pages` field in reverse publishing order according to the folder number, and ultimately it took stumbling on a forum post from earlier this year to find any reference to `num` as a sortBy option.

Since it seemed like a common enough case, I've added examples to two pages. In `docs/reference/panel/fields/pages` I've added several examples of how to use sortBy as an option, including caveats about how this disables sort handles/drag-and-drop reordering and that it doesn't affect actual sort order or template output (with a link to the Sorting cookbook entry on how to do that).

Then, in `docs/cookbook/content/sorting`, I've added a subhead to the end explaining that many panel fields can use sortBy: with modified syntax, one example, and a link to the Pages field entry to learn more.

I recognize this is a somewhat tricky thing to explain, so my caveats may not fit the style guide. Please let me know if I need to change anything.

I think it might also be good to add sortBy as a Field property on the Pages field, but I'm not comfortable doing that myself.

### DIFF
--- a/content/1_docs/2_cookbook/2_content/0_sorting/recipe.txt
+++ b/content/1_docs/2_cookbook/2_content/0_sorting/recipe.txt
@@ -230,3 +230,19 @@ $usera = $kirby->users()->sortBy('language', 'desc');
 ```
 
 For files and users (and all the other stuff we can sort) we could also go to extremes, but let's keep it dry and finish here.
+
+## Sorting Panel fields
+
+In blueprints, many Panel fields can also use sortBy with a slight variation on the standard syntax. Add `sortBy:` as a field option, and then list arguments in order without quotes or commas.
+
+For example, this would display a page's children in reverse publishing order: 
+
+```yaml
+fields:
+  posts:
+    label: Blog Posts
+    type: pages
+    sortBy: num desc
+```
+
+You can see more examples of how to use sortBy in the (link: docs/reference/panel/fields/pages text: Pages field).
--- a/content/1_docs/3_reference/3_panel/3_fields/0_pages/cheatsheet-article.txt
+++ b/content/1_docs/3_reference/3_panel/3_fields/0_pages/cheatsheet-article.txt
@@ -56,6 +56,42 @@ fields:
     max: 3
 ```
 
+## Sort order
+
+By default, the Pages field displays children in ascending publishing order from oldest to newest (`num asc`). In this state, each page has a drag handle to its left, allowing users to re-order pages by clicking and dragging.
+
+If you want to use a custom sort order when viewing in the panel (for example, to list blog posts in reverse post order, or by a 'date' field or any other page field), you can use the (link: docs/reference/objects/pages/sort-by text: sortBy) option. When used as a field option, function arguments should be listed in order, without quotes or commas. Please note this disables drag handles and drag-and-drop reordering in the Panel. You can still change the post order by clicking Change status and selecting a new position.
+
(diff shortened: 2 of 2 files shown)


texnixe:
Thanks a lot for your PR!

Adding `sortBy` in a pages field doesn't have any effect for me, and as far as I know that option doesn't exist?

Apart from that:

I think it doesn't make sense to add this single example to the cookbook recipe which is all about sorting in the frontend. In Panel blueprints, we have a lot of different places where we can use sorting, so putting one field example into the recipe doesn't make sense from my perspective.

Also, I think we should probably keep the sorting stuff for the reference much shorter.

Splendorr:
I can see these points. I’d like to ask where you think it might be best to put more detail about this function, or whether it already exists somewhere I couldn’t find. As someone who has worked with Kirby 2 and 3 on multiple projects over the last few years, I can usually find a way to approach using built-in functionality to solve a new problem without much difficulty. Or, I can determine that there’s no built-in way and go looking for other solutions.

However, the fact that I could even use sortBy as a blueprint option was not readily-apparent. That’s why I think including some mention of this in the Sorting Cookbook feels useful, even if it’s a mention that links to examples elsewhere. It is an instance of “Sorting.” Or the current Sorting entry could be re-titled Sorting Content in Templates and another, shorter entry created for Sorting in the Panel? 

Then it took multiple hours over several days of intermittent research to finally find someone on the forum explaining that “num” was the argument to use if I wanted to sort posts by folder number. I went back and couldn’t find any reference to this in the docs. So at least for me it seems like the docs assume I’ll know the possible arguments for sortBy.

So I edited the areas where I went looking for “how do I sort Pages in reverse order in the Panel,” but it’s okay if those weren’t the best places for it. I guess the things I’d like to improve are:

1. Discoverability of using sortBy as a blueprint option in the first place, and
2. Add a list of built-in arguments and/or usage examples.

Any guidance on how or whether I could do that best? Or, should I just simplify the examples in in the Pages field and maybe link to it from the Sorting cookbook page?

And, doing a bit more searching for where sortBy is explained, I see the Files panel entry includes sortBy as a field option, and a brief but helpful Sorting subsection. This is pretty much what I was looking for, although I really think “reverse order” for Pages is a common enough use case for people who are keeping a large set of Pages (as in a blog) and don’t want to click through many pages to get to their most recent entry. But I could model a new PR on the formatting in the Files panel page?

Thanks for your time and for helping me understand how to contribute most helpfully. :)

texnixe:
Hey @Splendorr, you are mentioning
> I see the Files panel entry includes sortBy as a field option
without a link to the page you are referring to. So I don't quite understand what you mean. As I said, I can't make your examples work.

The files and pages **sections** have a `sortBy` option, but the fields? What is the `Files panel entry`?

I think we have to sort that out before I can give you any tipps on where to add this, and I hope you understand that I can't accept a pull request for the description of an option that doesn't seem to work.

Splendorr:
@texnixe Oh goodness! I am using the Pages *section* but kept referring to the Pages *field.* Somehow in all of that searching, I never found my way to the Pages Section docs. Oops!

I think there's some way to improve the discoverability of this feature and its specific implementation, but the first problem was simply my being confused about that distinction.

I'll close this PR for now and think about where and whether there's some way to make this easier on someone else in the future. Sorry for wasting your time on this one.

texnixe:
The documentation for sorting a pages section is here: https://getkirby.com/docs/reference/panel/sections/pages#sorting

PR #659, opened 2019-08-14
fixed wrong var name

moritzebeling:


texnixe:
Thanks for spotting and fixing!

PR #660, opened 2019-08-14
loginPasswordless() method moved into if clause

afbora:
loginPasswordless() method moved into if clause, because of loginPasswordless() is void method, will not return a data.


lukasbestle:
Thanks a lot! I made some further improvements to the example.

PR #662, opened 2019-08-17
Add loader and template hooks plugin

afbora:
Add new [afbora](https://github.com/afbora) plugins:

- https://github.com/afbora/kirby-template-hooks
- https://github.com/afbora/kirby-loader


bastianallgeier:
I added them manually. Thanks for the PR :)

PR #671, opened 2019-08-29
Make it simple ...

Werbschaft:
The "File blueprint with two column layout" can be written much shorter, for me it was very confusing in the other way. If you're only using fields, you don't need to declare sections

### DIFF
--- a/content/1_docs/3_reference/3_panel/6_samples/0_file/cheatsheet-article.txt
+++ b/content/1_docs/3_reference/3_panel/6_samples/0_file/cheatsheet-article.txt
@@ -31,42 +31,34 @@ fields:
 title: Image
 
 columns:
-  main:
-    width: 1/2
-    sections:
-      content:
-        type: fields
-        fields:
-          caption:
-            label: Caption
-            type: textarea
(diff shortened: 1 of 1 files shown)


texnixe:
Thank you for the PR. I'm not sure we want this here. The purpose of the examples is not necessarily to show the simplest possible solution but rather something that is easily extendable, hence the sections. If we only add fields, we cannot add sections. The easy solutions are in our guide and the blueprint reference sections. Therefore I tend to leave it as is.

Werbschaft:
Sure, I understand! Only thought this makes sense. The new blueprints was for me the most hardest part when switching to Kirby 3 as they are hard to read and very complex.

texnixe:
Maybe we can add a comment that if you only want fields, you can simplify it and link to the guide. We are currently discussing ways of [simplifying blueprints in future versions](https://github.com/getkirby/ideas/issues/398).

Werbschaft:
Also possible!! For me it was hard to find out, that when only fields are needed, the whole Section Part isn't necessary. Same for the Columns when only working with fields

PR #674, opened 2019-08-31
Tiny cookbook fix

iskrisis:
Fix of this part that is confusing
https://getkirby.com/docs/cookbook/text/dates#the-date-field


texnixe:
Thank you, Florian. I made some more changes while at it…

iskrisis:
You shouldn't be workinh over weekends Sonja :)

texnixe:
But when, if not at the weekend?

PR #691, opened 2019-10-01
fix: remove margin-bottom on code blocks

pReya:
There is a unnecessary `margin-bottom` on code blocks. Since they have a black background-color, the margin shows as as a black bar below the scroll bar, which looks broken.

This is also redundant because there is already a `margin-bottom: 1.5rem` from the `.text figure` selector in `text.scss`.

Screenshot 2019-10-01 23 49 54

### DIFF
--- a/src/scss/components/code.scss
+++ b/src/scss/components/code.scss
@@ -43,7 +43,6 @@ h6 code,
   font-size: $font-size-code;
   letter-spacing: 0;
   line-height: ($line-height-code / $font-size-code);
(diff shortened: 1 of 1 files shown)


lukasbestle:
Thanks! Looks like none of us noticed it so far because we don't have the scrollbars set to be always visible.

pReya:
Neither do I. Saw it on examples, which have long lines that don't break (e.g. first listing on this page: https://getkirby.com/docs/cookbook/forms/basic-contact-form)

PR #694, opened 2019-10-05
Issue 693 disabled link in pagination

tomekpryjma:
SCSS linter errors resolved also.


lukasbestle:
Thanks for the PR! 👍
I have applied the fix manually because manually reviewing linter changes is quite an effort.

tomekpryjma:
Happy to help! Is there a preferred way of submitting PRs that include linter error fixes? In the future is it better to include them in the **Fix #issue-number** commit or a whole separate commit altogether?

lukasbestle:
The best way would be a separate PR for the linting fixes, as those are unrelated to the actual fix in the PR.

TBH it‘s even faster for us if you just open an issue like “there are CSS linting errors”, because running the linter ourselves is faster than reviewing a PR with hundreds of changed lines.

PR #714, opened 2019-10-20
Panel view overrides: add to Panel views reference

distantnative:
I wouldn't list the overrides as separate extension, but add the info to the general documentation of custom Panel views.


texnixe:
Yes, we can do that, but then we should change the intro, it doesn't really fit for the overrides.

distantnative:
@texnixe done

PR #717, opened 2019-10-22
Add note for those who use Composer to download a new copy of Kirby

jklue:
I use composer for new projects and wanted to also use it to download a fresh copy of Kirby 3 to update an old site. I used Composer to download in a directory next to the existing site, then followed steps 1, 2, and 3 in "Replacing core files and folders." The one important step I missed was bringing over the `/vendor` folder, which made it all work again.


lukasbestle:
If you run `composer require` directly in the site's directory, no further steps are needed.

To be honest, I think that this additional step is quite confusing if the reader doesn't know that it's about copying over a `kirby` folder from another installation.

jklue:
Wow! I didn't know you could just run `composer require` in the site directory. Would you need to download the composer.json first though? Agree that it's confusing otherwise.

lukasbestle:
No, you can run `composer require getkirby/cms` in any site directory and it will automatically set up a basic `composer.json` for you.

Of course it's always better to create an optimized `composer.json` beforehand (see our [docs](https://getkirby.com/docs/cookbook/setup/composer#adding-composer-to-an-existing-project)) to get better performance, but the most basic setup works fine as well.

PS: My first post was a bit misleading. `composer require` always needs the package name as an argument, in this case `getkirby/cms`. If you already have a `composer.json` with the dependencies defined, you can run `composer install` without any params to get Kirby installed automatically.

PR #722, opened 2019-10-28
Use absolute path for autoloader in custom setup

johannschopplich:
After setting up Kirby in a modern environment (`public` dir, …) and creating a custom PHP router, calling the `public/index.php` results in `autoloader.php not found`, because it's called from a relative path.

In the [Shared storage folder](https://getkirby.com/docs/guide/configuration#custom-url-setup) example below the [Public folder setup](https://getkirby.com/docs/guide/configuration#custom-folder-setup__public-folder-setup), the autoloader is also called via an absolute path.


lukasbestle:
Thanks! 👍

Issue #646, opened 2019-08-06
Document 3.2.3 changes


texnixe:
- [x] Support for sorting constants in the sortBy option in sections (#1913)
- [x] Optional content lock for virtual pages (#1539

distantnative:
✅

PR #730, opened 2019-10-31
Guide for content locking of virtual pages #646

distantnative:


texnixe (approved):
We should probably go into more depth and provide an example in a cookbook recipe later on, but this is a good start.

distantnative:
@texnixe Yes, but to be honest I think implementing content locking for API virtual pages or so can turn reaaaaally complicated and advanced. I think for now it's fine to leave those really advanced users with the pointers.

PR #744, opened 2019-11-11
Fix missing since info

texnixe:


distantnative:
But I think we should actually also fix this in the core code

texnixe:
There are two different issues. One is missing since information in the core code, the other issue that I'm trying to fix with this PR is `since` info in pages being ignored.

PR #754, opened 2019-11-21
JPG and PNG files optimized

afbora:
JPG and PNG files optimized with [TinyJPG](https://tinyjpg.com/) tool

When I started optimizing, I didn't think it would save that much 😲 
But I see that it has been a serious resource saving and will be reflected in site speed and performance 🚀 

**Before**: ~153 MB *
**Optimized Files**: 396
**After**: ~77 MB *
**Saved**: ~76 MB / ~49%

Optimized files details: [OptimizeLogs.zip](https://github.com/getkirby/getkirby.com/files/3873900/OptimizeLogs.zip)

_* Whole getkirby.com size_


distantnative:
Some of the images don't seem to be as high-res anymore after this.
But it might still be fine for the context they are used in. What do you think, @texnixe?

texnixe:
I haven't had a chance to look at the result yet, so I don't have an opionion atm. We should still make sure that images are crisp enough, even in the documentation pages.

distantnative:
Actually I have to correct myself: compared some images in the reference, on the front page, on the plugin community page... no visible difference for my eye.

afbora:
@distantnative Yes that is right. As I remember and check again, only the resolution of the following 2 images has changed. These images are in the kosmos and are about 9MB in size. Others have the same resolution and almost (99%) the same quality.

1. https://github.com/getkirby/getkirby.com/blob/348d49944c03194fc7670b368e87882c2ca19761/content/kosmos/20190301_32/fair-climate-fund.png

2. https://github.com/getkirby/getkirby.com/blob/7c7532670204a5404cce327e29470036522fdcc3/content/kosmos/20190705_36/bruno-moinard-editions.png

You see the statistics below. Why don't we use resize or crop methods here?

image (5)
image (6)

texnixe:
Then lets merge this and save energy and bandwidth 🙂

afbora:
@texnixe Still, I'd like you to check it out as a second eye.

texnixe:
I trust your eyes better than mine, guys.

PR #765, opened 2019-11-26
Syntax note

jklue:
I didn't realize the `pages/` prefix was needed for the blueprint value. Thought a note to that effect may help someone else (and me when I come back to this topic).


distantnative:
I agree that we might need a notice there. But now it would be too restrictive. It needs the `pages/` prefix if you want to register a page blueprint - same as the normal folder structure. But if you want to register a user blueprint for example, it needs a different prefix

jklue:
Good point—hadn't considered those use cases. I've updated the note. Is it better?

PR #766, opened 2019-11-26
Update cookbook code example about securing JSON 

johannschopplich:
Since the `$request->ajax()` method is removed, the cookbook article about generating JSON should be updated as well. Rather than deleting the entire paragraph, I think it's worthwhile to keep it.

You may find my idea appealing, or you may not. 🤷‍♂️ What do you think?


bastianallgeier (approved):

bastianallgeier:
Great update! Sorry that it took a while to review it

PR #767, opened 2019-11-26
Missing { after if conditional

paulstone:
In one of the examples, the opening `{` is missing just after the conditional.

```
if($autoPublish) 
    $page->changeStatus('listed');
}
```

should be:

```
if($autoPublish) {
    $page->changeStatus('listed');
}
```


texnixe:
Thanks for spotting and fixing 🙂 .

PR #769, opened 2019-11-30
Changes to features page

distantnative:


bastianallgeier:
I included your changes in my branch, but kept the URLs without /product

PR #770, opened 2019-12-01
Some more feature page changes

texnixe:


site/templates/features.php
```diff
@@ -9,7 +9,7 @@
       <section class="features-section">
         <?= snippet('features/header', [
           'heading' => 'The Panel',
-          'subheading' => 'A control room that feels like home'
+          'subheading' => 'A control room where you will instantly feel at home'
```
distantnative, line 12: Personally, here I like Bastian's original line better. Kurz und knack :D

site/templates/features.php
```diff
@@ -21,23 +21,23 @@
...
-              Prepare your content - pages, articles, events, products - and only publish them when everything is in place with the click of a button. Send preview links to others so they can review your drafts even before they go live.
+              Prepare your content—pages, articles, events, products—and only publish them when everything is in place with the click of a button. Send preview links to others so they can review your drafts before they go live.
             '
           ]) ?>
 
           <?= snippet('features/text', [
             'heading' => 'Drag & Drop sorting',
             'text' => '
-              Sorting shouldn\'t be complicated. Just pick your pages or files up and drop them where you want them to be. Why should it be anymore complicated than that?
+              Sorting pages or files is a breeze. Just pick them up and drop them where you want them to be. It shouldn\'t be more complicated than that?
```
distantnative, line 40: Keeping the question mark at the end intended?

site/templates/features.php
```diff
@@ -365,11 +365,11 @@
           ]) ?>
           <?= snippet('features/text', [
             'heading' => 'KirbyTags',
-            'text'    => 'When Markdown just isn\'t enough, add your own KirbyTags to simplify plain text formatting and injection of complex custom components into long-form text.'
+            'text'    => 'When Markdown isn\'t enough, add your own KirbyTags to simplify plain text formatting and to inject complex custom components into long-form text.'
           ]) ?>
           <?= snippet('features/text', [
             'heading' => 'Caching',
-            'text'    => 'Kirby comes with mighty caching on board. Not fitting for your project? Add the cache driver on your choice and reduce page loads to the blink of an eye.'
+            'text'    => 'Kirby comes with mighty caching on board. Not fitting for your project? Add your cache driver of choice and reduce page loads in the blink of an eye.'
```
distantnative, line 372: Here I meant that the page load only takes a blink of an eye. Not the implementation of the cache driver. That would be a bit exaggerated.

distantnative:
Great! I really like the changes. Only have a few comments/questions.

site/templates/features.php
texnixe: Maybe we can find something better, but feels like home sounds strange to me in combination with a control room.

site/templates/features.php
texnixe: I actually think the whole question is redundant and should be replaced with something better. I didn't have a good idea yet.

site/templates/features.php
texnixe: I've never seen the prepositon "to" in combination with "blink of an eye". Maybe "reduce page load massively".

bastianallgeier: How about "reduce page loads significantly"

site/templates/features.php
bastianallgeier: Maybe we should remove the "control room" part entirely and make it more human friendly.

bastianallgeier:
All changes are now in my branch again, but without the /product URLs. Let's move on from there.

site/templates/features.php
texnixe: That sounds much better.

PR #839, opened 2020-01-16
Create new “Resources” section

distantnative:


bastianallgeier:
Let's close this as we already handle it in the new version.

distantnative:
🙌

PR #842, opened 2020-01-22
"categories" question

luxuryluke:
On line 190 in the template code sample, shouldn't this:
`…->categories()->…`
be this:
`…->category()->…`
…since it's referencing a field defined as `category` (from line 176)?

If this is not true, then there is some pluralization of field names that I'm not aware of (and that's very possible). ;)


texnixe:
Good catch! Thanks!

PR #843, opened 2020-01-22
Use accessible markup in example

sebastiangreger:
As per https://github.com/scottaohara/a11y_breadcrumbs this would be the preferred markup for a breadcrumb navigation.


texnixe:
We should probably also remove the role attribute...

texnixe:
Thank you, @sebastiangreger !

PR #847, opened 2020-01-27
Add Colors and Tree Method plugins

nilshoerrmann:
This updates our plugin informations, adding new pages for `kirby-colors` and `kirby-tree-methods`.

### DIFF
--- a/content/plugins/hananils/colors/plugin.txt
+++ b/content/plugins/hananils/colors/plugin.txt
@@ -1,0 +1,13 @@
+Title: Colors
+
+----
+
+Repository: https://github.com/hananils/kirby-colors
+
+----
+
+Category: panel
+
+----
+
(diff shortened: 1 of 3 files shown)


bastianallgeier (approved):

bastianallgeier:
It would be lovely to have a nice screenshot for the colors plugin. The ones in the Github repo are a bit too small I think.

nilshoerrmann:
What kind of bigger screenshot do you imagine? And how do I provide it best?

bastianallgeier:
The format should be similar to this: https://getkirby.com/plugins/sylvainjule/autofocus

Maybe you can arrange a couple different field setups in a nice grid (3 x 3 or 2 x 2 color fields) and make a screenshot of that. 

Kirby will pick up everything that is called screenshot.(png|gif|jpg) in the same folder as the plugin.txt

nilshoerrmann:
Okay, I'll try to take a look into that next week.

PR #858, opened 2020-02-09
Merge with 3.3.4: Language locales: Document string keys

lukasbestle:


afbora:
This page also contains:

https://getkirby.com/docs/reference/system/options/locale

lukasbestle:
In the config file it’s alright as that one is never touched by Kirby.

But now that I think about it: Maybe we should actually keep the constants in the docs. If added manually, the bug does not occur. It’s only an issue with the constant ints.

PR #859, opened 2020-02-09
Add first panel field cookbook recipe

texnixe:
Needs new screenshots after review.


bnomei:
great draft. i love the fact you explain plain js and vue. makes it so easy to get started with vue then.

PR #875, opened 2020-02-15
Add/extends filter methods for filterBy() in files and users

texnixe:


distantnative:
I feel like we should rather put this into a snippet instead having duplicate info

texnixe:
I agree. But since it should also be in the collection and structure classes and that needs some restructuring, this was a first step.

PR #879, opened 2020-02-20
Fix migration script file name in migrate users documentation

minimalweb:
Should be `migrate.php` instead of `migration.php` to be consistent (otherwise if following the instructions exactly, visiting the URL wouldn't work)


lukasbestle:
Thanks!

PR #883, opened 2020-02-26
Can you create a Title field?

luxuryluke:
Line: 28 & 54
Doesn't the Title field conflict with the title of the page?
Perhaps this should be changed to a different example since `title` is reserved for that.

You can ignore this if I've misunderstood.


texnixe:
Yes, it should be removed. But we probably want to rename the draft field as well as it conflicts with the `$page->draft()` method.

texnixe:
Thanks for spotting and your PR!

PR #892, opened 2020-03-06
Add example for $pages->group() method

sebastiangreger:
Great example by @texnixe from https://forum.getkirby.com/t/group-articles-by-the-first-character-of-their-title-name/17521 - very valuable to grasp how a callback function works


distantnative:
Thanks!

Issue #791, opened 2020-01-08
Best practices for plugins


texnixe:
Apart from the composer etc. stuff
- structure
- tagging
- good description of what the plugin actually does
- who needs it?
- versioning
- naming conventions

GiantCrocodile:
+ info about mentioning the license under which it is published and if it's free/commercial

PR #913, opened 2020-03-19
Add plugin best practices

texnixe:
#791


lukasbestle (approved):
This is probably a recipe that can and should be extended later with feedback from other plugin devs from the community. I fully agree with everything that is in there right now.

PR #936, opened 2020-04-19
changes images() to files() in filter recipe

flokosiol:
I thing there are two `images()` method in this recipe where `files()` should be used instead.


texnixe:
Thanks, Flo!

flokosiol:
👍 Typo included (in my comment) 🙈😂

PR #937, opened 2020-04-20
fix: typo progamatically → programatically

jonaskuske:
:)


texnixe:
There was still a second m missing... Thank you!

jonaskuske:
Whoops, right!

I had actually googled "programatically" and the first result was a wiktionary entry for that exact word so I went with it – just clicked on the result and the page simply reads "Misspelling of programmatically". Wow. 😄

PR #939, opened 2020-04-21
fix: typos and broken link

jonaskuske:
Didn't clone the entire repo to actually test it, but I hope that just adding the colon after `link` will fix that link?


lukasbestle:
Yes, thanks! 👍

PR #947, opened 2020-04-29
First stab at a Backups recipe

sylvainjule:
I have merged the recipe from Notion in a txt file with all the images / videos.

Of course feel free to emendate / re-phrase anything.

Also the recipe has a few videos, I haven't found any recipe with one yet and am not sure how you'd like to integrate them, so I added the mp4 file and referenced their filenames where they should appear.


texnixe:
Hey @sylvainjule, sorry I didn't find the time to look into your recipe earlier. I was trying to follow along but at some time lost track, because of the left-out parts in the code. Of course, it wouldn't make much sense to repeat everything all the time, but I wonder if we shouldn't provide a wrap-up after each section with the complete code? Maybe we need hideable sections for such purposes.

Using HTML comments in the template parts could also help. I'll send a PR when I'm through with it.

Apart from these little things, great recipe!

We also need a way to include the videos...but that's another story.

sylvainjule:
Yes I struggled a little with the page length the recipe would have if I repeated the whole code at each change. 

I went for explaining things bit by bit rather than big chunks of commented code, thought it would be easier to follow but perhaps it isn't. I think having a code symmary at the end of each section would be really nice, but indeed it could really help if they could be toggled because I feel like the page would be sooooooo long with those.

distantnative:
I think this is still blocked (aside from changes) due to no video embeds, correct?

texnixe:
Which is unfortunate because this is a really good recipe that's been hanging around for ages.

distantnative:
I created a very minimal video tag. However given the width, the videos are quite small inline, especially those with already a narrow crop:

Screen Shot 2021-01-05 at 23 35 09

What do you both think?

texnixe:
Probably doesn't matter as users can always download them if necessary.

The available width is a general problem, not only for videos, but also for overview screenshots.

distantnative:
I pushed it to a branch in this repo since I can't alter the fork: https://github.com/getkirby/getkirby.com/tree/sylvainjule/backups-recipe

Let me know what you both think, if this works for this recipe. We can still improve the KirbyTag when refactoring all our custom tags.

sylvainjule:
I think it works fine indeed. 👍

distantnative:
@texnixe what do you think?

would it be ready then to be merged? content-wise?

texnixe:
@distantnative I once made some suggestions for changes, but don't know what happened to them... Doesn't look like they were merged? Maybe @sylvainjule  wasn't d`accord.

texnixe:
But as far as I remember, there were no issues at all content-wise.

afbora:
I also integrated the new video tag into the branch @distantnative is committing to:
https://github.com/getkirby/getkirby.com/tree/sylvainjule/backups-recipe

⚠️ Rebased the commits from the PR against develop.

afbora:
Moved to #1190

PR #952, opened 2020-05-08
Monolithic plugin recipe

yandodov:
Adding a recipe about a plugin setup where a single repository holds both your plugin and a site where you can test it. The green light for this was given to me by @bastianallgeier and @texnixe


texnixe:
Cool, thanks a lot!

PR #967, opened 2020-05-29
Use require instead of require_once

fabianmichael:
Although the plugin file should only be loaded once, I think using `require` makes more sense at this point, because loading the file for a second time would not register the same plugin data, as on first load.


lukasbestle:
Absolutely, thanks! 👍

PR #969, opened 2020-06-02
Security page

lukasbestle:
This will also be used by the planned update checker. Also see https://github.com/getkirby/kirby/pull/2627#issuecomment-637276051.

### DIFF
--- a/content/1_docs/1_guide/18_security/guide.txt
+++ b/content/1_docs/1_guide/18_security/guide.txt
@@ -11,7 +11,9 @@ Text:
 
 ## Always update your Kirby installation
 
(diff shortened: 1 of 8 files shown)


distantnative:
I like it!
But wondering about the „supported“ phrasing being tied to security incidents. The v3 supported/recommended version would always be the latest bugfix release - no matter if security fix or other bugfix, won’t it?

lukasbestle:
@distantnative I agree and I‘ve done it like that in the content file. But now I see that I forgot to change the description in the JSON template.

Or do you mean the headline/JSON key in general? If so: Any ideas what would be a better term?

distantnative:
Don’t really know a better one 😂 I think it’s fine

PR #972, opened 2020-06-12
feat: adds webhook-field to plugin directory

pju-:
This has been in beta for ages, but I have made it installable via composer now, so it should be ready to go :)

Thanks for your work!


lukasbestle:
Thanks! 👍

PR #975, opened 2020-06-14
Add trailing slash recipe

texnixe:
It would probably make sense to link to this mini recipe from the guide or so.


lukasbestle (approved):

lukasbestle:
Thanks! 👍

PR #978, opened 2020-06-19
Replace non-existing sort field argument with collection, fix #976

### LINKED ISSUE #976: Error in next, prev, nextListed, prevListed docs
The expected argument for those methods is a Collection, while the examples mention sort arguments: [/docs/reference/objects/page/next#example__using-sort-field](https://getkirby.com/docs/reference/objects/page/next#example__using-sort-field)

```php
// doesn't work
if($next = $page->next('date', 'desc')):

// works
if($next = $page->next($page->siblings()->sortBy('date', 'desc')):
```

texnixe:


content/1_docs/3_reference/4_objects/0_page/0_has-next-listed/method.txt
```diff
@@ -8,10 +8,11 @@ echo $page->hasNextListed(); //will echo 1 if true, nothing if false.
 ?>
 ```
 
-### Using sort field
+### With collection as argument
 
 ```php
 <?php
-echo $page->hasNextListed('date', 'desc'); //will echo 1 if true, nothing if false.
+$collection = $page->siblings(false)->listed()->sortBy('date', 'desc');
```
lukasbestle: Does this work with `->siblings(false)`? The `next`/`prev` methods need to find the position of the current `$page` in the collection to move from there. So if the `$page` itself is not part of the collection, I wonder how it will work. But I haven't tested this.

Unrelated change: The `will echo 1 if true, nothing if false.` comments are quite confusing. The `has*` methods actually return a boolean, so it would be better if the examples used `var_dump()`, which will clearly display the result and will make the comment unnecessary.

texnixe: Good question, I'll double-check, but I think you are absolutely right and I just didn't think properly.

lukasbestle (approved):

PR #986, opened 2020-06-30
Update subpage builder recipe

texnixe:


content/1_docs/2_cookbook/7_extensions/0_subpage-builder/recipe.txt
```diff
@@ -148,3 +148,62 @@ function buildPageTree($page) {
...
+
+```php "/site/config/config.php"
+'hooks' => [
+    'route:after' => function ($route, $path, $method) {
+        $uid = explode('/', $path);
+        $uid = end($uid);
+        $uid = str_replace('+', '/', $uid);
+        $page = kirby()->page($uid);
+
+        if ($page->children()->isEmpty()) {
```
medienbaecker, line 189: I think this could lead to an error when `$page` is null/doesn't exist. Maybe add a `$page` check to the condition?

texnixe: How could I miss that, preaching that all the time in the forum 🙈 ?

PR #991, opened 2020-07-09
New cookbook recipe: Disable page cache for logged-in users

sebastiangreger:
After sheer endless experimentation with this (some documented [here](https://github.com/getkirby/ideas/issues/40) and [here](https://forum.getkirby.com/t/disable-cache-for-logged-in-users/14197/7)), I finally believe to have found a working solution that is not even a hack.

Wrote this as my first cookbook recipe, as this seems to be a common itch and I considered it worthwhile to document some of the learnings from the process for others to learn from.

Happy to iterate if you find anything that could be improved :)

### DIFF
--- a/content/1_docs/2_cookbook/9_setup/0_selective-page-cache/recipe.txt
+++ b/content/1_docs/2_cookbook/9_setup/0_selective-page-cache/recipe.txt
@@ -56,7 +56,7 @@ While it may appear obvious to replace `'active' => true` with `'active' => !kir
 
 It would be possible to move the test for a logged-in user into the (link: docs/reference/system/options/ready text: `ready` option), executed after Kirby is completely initiated, but another problem emerges with modifying the `active` setting:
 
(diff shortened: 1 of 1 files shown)


content/1_docs/2_cookbook/9_setup/0_selective-page-cache/recipe.txt
```diff
@@ -0,0 +1,117 @@
...
+return [
+    'cache' => [
+        'pages' => [
+            'active' => true
+        ]
+    ]
+];
+```
+
+Kirby also takes care of emptying the cache whenever content is changed through the panel so that no outdated content is served to users.
```
texnixe: The Panel needs to start with a capital P ;)

sebastiangreger: Fixed.

lukasbestle (approved):
That's a really awesome and simple solution. I also really like the detailed and deep explanations. 👍

Thank you very much for writing the recipe! ❤️

PR #1001, opened 2020-07-23
Fixed a typo in the toggle input documentation

skyeewers:
This simple typo caused the example to not produce any results when copy and pasted into a Kirby plugin


distantnative:
Thanks!

PR #1008, opened 2020-08-12
Add missing footnotes plugin

sebastiangreger:
Unless omitted on purpose, this plugin by @sylvainjule is missing from the plugin directory.


lukasbestle:
Thank you! Seems like we forgot to add it. 👍

PR #1023, opened 2020-09-08
add clarity to preview image query by field

jklue:
from https://forum.getkirby.com/t/panel-cover-image/15280/2?u=lukehatfield


texnixe:
Thank you, @lukehatfield!

PR #1026, opened 2020-09-11
clarify template use of collections

luxuryluke:
I just think it would be helpful to clarify the use of collections in a template.
By adding these lines to a template, does it output them? Does it only set them up as referenceable arrays? etc.


distantnative:
Could you please open an issue instead of a PR that would change the website content, adding your question to it?

PR #1028, opened 2020-09-12
Business hours example is incorrect

steve65140:
I think the original author copied and pasted the third example `echo I18n::template('file.max.error', null, [` when it should be `echo I18n::template('business.hours', null, [`.


texnixe:
Thank you!

PR #1031, opened 2020-09-15
update ValidationRules: min to minLength, max to maxLength

dennisbaum:
Change the validation-rules from min to minLength. otherwise i can put 4 in the name and it validates without error, but i guess it should check for the lengths of the characters, not the numeric value... same for max/maxLength


bastianallgeier:
Thank you for the fix!

PR #1036, opened 2020-09-18
Update recipe.txt

fritzlaszlo:
$subMenuItems is undefined here. I think it should be $subMenu. Just did a quick test and I think it works properly like this.


texnixe:
Thank you very much!

PR #1041, opened 2020-09-25
Fix link to PHP `strftime` function in `toDate` docs

jhogervorst:
The page used to link to the PHP function `strtotime`, which takes a different format than `strftime`.


distantnative:
👍 thanks

PR #1043, opened 2020-09-29
Add mailhog recipe

texnixe:
How to use MailHog to test sending email with Kirby


bastianallgeier (approved):
Woah, this is so cool! I urgently needed such a guide myself <3

PR #1044, opened 2020-10-01
Fix typos

medienbaecker:


texnixe:
Thanks, Thomas! I really love how everyone contributes to making our docs and Kirby better!

PR #1050, opened 2020-10-08
update code sample to match image

jklue:
This makes the sample code match the screenshot above it. Before it seemed like the `label:` was confused for the `text:`? Also added explicit "no" and "yes" options to match screenshot.


texnixe:
Absolutely! Thank you!

PR #1051, opened 2020-10-16
Update recipe.txt

janstieler:
Add Exception $error output to $alert.

### DIFF
--- a/content/1_docs/2_cookbook/4_forms/0_basic-contact-form/recipe.txt
+++ b/content/1_docs/2_cookbook/4_forms/0_basic-contact-form/recipe.txt
@@ -146,7 +146,7 @@ return function($kirby, $pages, $page) {
                 ]);
 
             } catch (Exception $error) {
-                $alert['error'] = "The form could not be sent";
+                $alert['error'] = "The form could not be sent:" . $error;
             }
 
             // no exception occurred, let's send a success message


content/1_docs/2_cookbook/4_forms/0_basic-contact-form/recipe.txt
```diff
@@ -146,7 +146,7 @@ return function($kirby, $pages, $page) {
                 ]);
 
             } catch (Exception $error) {
-                $alert['error'] = "The form could not be sent";
+                $alert['error'] = "The form could not be sent:" . $error;
```
afbora, line 149:
```suggestion
                $alert['error'] = "The form could not be sent:" . $error->getMessage();
```

janstieler: Your right this is a nicer output.

texnixe:
Not so sure if this is the kind of information you would want to show to a site visitor and therefore we only output a general error message.

janstieler:
Hello Sonja,
okay that is a point so I would add a if-clause which is query the debug config:
```
 if(option('debug')):
    $alert['error'] = 'The form could not be sent: <strong>' . $error->getMessage() . '</strong>';
 else:
     $alert['error'] = 'The form could not be sent!';
endif;
```
The reason for this commit is that i had lost a couple of Time because I don't see that this example hat no $error-output.
Which is frustrating if you had copied this code .

Greetings
PS: Should I make a new commit with this or?

texnixe:
Hey Jan, yes please!

PR #1058, opened 2020-11-05
Update recipe.txt

fritzlaszlo:
Added a missing $.
Deleted ('photography'), which was not in line with the other examples in this cookbook and hence could confuse readers.


texnixe:
Makes sense. Thank you!

PR #1060, opened 2020-11-08
Texnixe/oop

texnixe:


bastianallgeier:
@texnixe why did you close the PR?

texnixe:
@bastianallgeier Because I first wanted to proof-read it again and then I thought I could add a thought or two here and there. But feedback welcome!

bastianallgeier:
Ok, no problem. I will wait a bit more in this case.

texnixe:
@bastianallgeier Do you think it's useful?

bastianallgeier:
Yes, absolutely! I think our docs would really stand out even more if we give PHP beginners more guidance.

texnixe:
That was the idea. Another one I've started is using Xdebug.

PR #1061, opened 2020-11-09
more consistency in controller update

benhatfield:
In the example, it was unclear that we still need the $site, $pages, and $page variables as arguments/parameters in a controller when we upgrade to Kirby 3. Keeping those as parameters highlights that just the $args variable was removed.

### DIFF
--- a/content/1_docs/2_cookbook/9_setup/0_migrate-site/recipe.txt
+++ b/content/1_docs/2_cookbook/9_setup/0_migrate-site/recipe.txt
@@ -264,7 +264,7 @@ to
 ```php
 <?php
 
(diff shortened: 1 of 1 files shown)


texnixe:
This part of the migration guide is no longer up to date: https://getkirby.com/docs/guide/templates/controllers#arguments-from-page-render-in-route

Also, you only need to pass the arguments you really use within your controller in any order (whereas in Kirby 2 you always had to pass all arguments ($page, $site, $kirby, $pages in the right order).

distantnative:
Closing in favor of https://github.com/getkirby/getkirby.com/pull/1141

PR #1063, opened 2020-11-10
Update guide.txt

danielgroner:
Changed the three parts in the describing texts saying "should be redirected to the..." from "...error page" to "...index page", since the RewriteRules for Blocked folders and files redirect to index.php / homepage, not the error page.


bastianallgeier:
It redirects the request to the index.php but the URL is not redirected to / and thus you will always land on Kirby's error page. I think our description is therefor more correct to be honest.

danielgroner:
Ah I see, I actually checked it on a website I built, there it always directed to the homepage. But on another one it directed to the error page as supposed to. So your description indeed is the correct one. Sorry :)

bastianallgeier:
Maybe you still have an old htaccess in the other project? I think we actually redirected to the home page in earlier versions.

PR #1064, opened 2020-11-14
Add OOP recipe

texnixe:


content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+If you want to follow along with this recipe, you need the following:
+
+- A web server (ideally on your (link: docs/cookbook/setup/development-environment text: local machine))
+- A current version of PHP
+- A Kirby installation ((link: docs/guide/quickstart text: Starterkit or Plainkit))
+- A code/text editor of your choice
+
+## What is OOP?
+
+Object Oriented Programming is a programming paradigm based on the concept of objects that have certain properties and methods that allow interacting with these objects. The properties and methods of such an object are described in a class, which acts as the model or blueprint for such objects, whereas an object is an instance of this class, i.e. a specific embodiment of the class.
```
lukasbestle: I know there's an example right after this, but I feel this is can be too abstract for readers without OOP experience. It could be that readers stop reading because they didn't understand the explanation.

What if we first had a relatable real-world example and then the abstract explanation afterwards. Something like:

> Object Oriented Programming (OOP) is all about objects, their properties and their methods.
>
> Let's first think about a real-world example: An object could be an apple. Its properties are the variety, color, weight and size. And its methods are the actions you could do with the apple, e.g. you can pick it up, give it to someone or eat it.
>
> OOP uses the same paradigms for data structures in a program. The properties and methods of an object like the apple are described in a class, which acts as the model or blueprint for every apple you might have in your application, whereas an object is an instance of this class, i.e. a specific embodiment of the class (the specific apple you can hold in your hand).

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+In a larger software program, you typically have multiple classes whose objects can interact with each other.
```
lukasbestle:
```suggestion
In a larger software program, you typically have multiple classes whose objects can interact with each other (maybe an apple, a tree and a farmer).
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+### A simple class example
+
+Let's examine a very simple example to get a basic idea. Here we define a class called `Book` with some properties that should apply to all objects of this class, and methods to work with objects of this class.
```
lukasbestle:
```suggestion
Now let's take a look at a PHP example of OOP. Here we define a class called `Book` with some properties that should apply to all objects of this class, and methods to work with objects of this class.
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+```php "book.php"
+<?php
+class Book {
```
lukasbestle: It would be good to have the class set up according to the [PSR-2 standard](https://www.php-fig.org/psr/psr-2/):

```suggestion
class Book
{
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+    private $title;
+    private $author;
+    private $isbn;
+    private $price;
+    private $publisher;
```
lukasbestle: I'd make these `protected` as the use-case for `private` properties is limited.

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+    public function __construct( string $title, string $author) {
```
lukasbestle:
```suggestion
    public function __construct(string $title, string $author)
    {
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+    
+    public function getIsbn(): string {
+        return $this->isbn;
+    }
+    
+    public function setTitle(string $title): void {
+        $this->title = $title;;
+    }
+
+    public function setAuthor( string $author): void {
```
lukasbestle:
```suggestion
    public function setAuthor(string $author): void
    {
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+## Describing a class
+
+First of all, we define our new `Book` class with the keyword `class`. Inside our class, we define the properties of this class, like the title, the author, the publisher etc. Of course, this book can have many more properties, but we keep it basic.
+
+### Properties and methods
+
+Class properties (variables and constants) and also the methods can have different states of visibility: private, protected or public. Properties of a class should be private or protected to keep control over what can be assigned to a property and what not.
+
+To access these properties, we provide public so called getters that return the value of a property (`getPrice()`, `getTitle()` etc.) and setters which can be used to assign a value to the property of an object (`setPrice()`, `setTitle()` etc.). Getters and setters are often prefixed with `get-` and `set-` to make it obvious what these methods do, but this is not mandatory ( and you will not find a single method named that way in Kirby's core).
```
lukasbestle:
```suggestion
To access these properties, we provide public so called getters that return the value of a property (`getPrice()`, `getTitle()` etc.) and setters, which can be used to assign a value to the property of an object (`setPrice()`, `setTitle()` etc.). Getters and setters are often prefixed with `get-` and `set-` to make it obvious what these methods do, but this is not mandatory (and you will not find a single method named that way in Kirby's core).
```

We do prefix our setters with `set` BTW.

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+So what's all this?
+
+## Describing a class
+
+First of all, we define our new `Book` class with the keyword `class`. Inside our class, we define the properties of this class, like the title, the author, the publisher etc. Of course, this book can have many more properties, but we keep it basic.
+
+### Properties and methods
+
+Class properties (variables and constants) and also the methods can have different states of visibility: private, protected or public. Properties of a class should be private or protected to keep control over what can be assigned to a property and what not.
```
lukasbestle, line 117: I'd write something about the difference between `private` and `protected` here as it's otherwise confusing:

> Both private and protected properties cannot be accessed from the outside of the class, private properties are even more limited in access. In most cases, protected properties are fine. You can read more about the difference below.

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+In a new folder in your webroot, create a file called `book.php` with the class definition from above. Next to it, create an `index.php` file with the following content:
+
+```php "index.php"
+<?php
+// make PHP display errors on screen
+ini_set('display_errors', 1);
+require_once __DIR__ . '/book.php';
+
+// we create a new book object, by setting a title and an author as required by the constructor
+$book = new Book( 'The Hitchhiker\'s Guide to the Galaxy', 'Douglas Adams');
```
lukasbestle:
```suggestion
$book = new Book('The Hitchhiker\'s Guide to the Galaxy', 'Douglas Adams');
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+In a new folder in your webroot, create a file called `book.php` with the class definition from above. Next to it, create an `index.php` file with the following content:
+
+```php "index.php"
+<?php
+// make PHP display errors on screen
+ini_set('display_errors', 1);
+require_once __DIR__ . '/book.php';
+
+// we create a new book object, by setting a title and an author as required by the constructor
```
lukasbestle, line 141:
```suggestion
// we create a new book object, by setting a title and an author as required by the constructor
// (the `__construct()` method that creates new objects of the class)
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+// we have declared our properties as private, so we cannot access them directly
+echo $book->author; // this line will throw an error ( Fatal error: Uncaught Error: Cannot access private property Book::$author )
+*/
+
+// but we can use the getters to get these properties from the book object
+// Note that you have to comment the above error line before you will see the results of this code
+echo $book->getTitle();
+echo $book->getAuthor();
+
+// the price is not yet set, so this will no print anything
```
lukasbestle:
```suggestion
// the price is not yet set, so this will not print anything
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+/*
+// Let's create another error
+$book = null;
+echo $book->getAuthor(); // Fatal error: Uncaught Error: Call to a member function getAuthor() on null
+*/
+
+/*
+$book2 = new Book('Beloved', 'Toni Morrison')
+echo $book->type(); // Fatal error: Uncaught Error: Call to undefined method Book::type()
```
lukasbestle:
```suggestion
echo $book2->type(); // Fatal error: Uncaught Error: Call to undefined method Book::type()
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+```
+Fatal error: Uncaught Error: Call to a member function getAuthor() on null
+```
+
+will probably look familiar if you have worked with Kirby for a while.
+
+What's the issue? We have declared the `$book` variable, but instead of instantiating an object of the book class, we have assigned `null` to the variable. The method we try to call (`getAuthor()`), however, is not some function you can use anywhere at will, it is a member function (or method) of the `Book` class and can therefore *only* be used with objects of this class. The same error (with a slightly different error message) would appear if `$book` was a string, a boolean or an integer. Try it out.
+
+On the other hand, if you have a `book` object, but try to use a method that is not defined in the class, you will also get an error, this time
```
lukasbestle:
```suggestion
On the other hand, if you have a `Book` object, but try to use a method that is not defined in the class, you will also get an error, this time:
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+You can create as many book objects from this class as you like, by providing the minimum properties required to define a book (in this case the title and the author).
+
+<info>
+In our example above, we have also introduced type declarations for parameters and return types. This concept is not limited to OOP, but makes sure that we get the correct input values and that our methods return what they are expected to return. An error is thrown when the type declarations are not met.
+</info>
+
+### The constructor
+
+In the example above, we declared a `__constructor()` method. This method is optional, but allows us to initialize an object's properties (or the properties we regard as essential) upon creation of the object. In our example, we have to pass the title and the author parameters both as strings when creating a new instance of the class and cannot just create a new Book with `new Book()`. It is also possible to set default values for certain parameters (which doesn't really make sense for our books). But all of this totally depends on the class and the objectives.
```
lukasbestle:
```suggestion
In the example above, we declared a `__construct()` method. This method is optional, but allows us to initialize an object's properties (or the properties we regard as essential) upon creation of the object. In our example, we have to pass the title and the author parameters both as strings when creating a new instance of the class and cannot just create a new Book with `new Book()`. It is also possible to set default values for certain parameters (which doesn't really make sense for our books). But all of this totally depends on the class and the objectives.
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+Looks a bit different than our mini example above, doesn't it? It's not only very much longer but it doesn't even start with the class definition. Let's go through what this stuff at the top means.
+
+### Namespacing
+The first line after the opening PHP tag reads:
+
+```php "/kirby/src/Cms/Page.php"
+namespace Kirby\Cms;
+```
+
+Here, we declare a (link: https://www.php.net/manual/en/language.namespaces.rationale.php text: namespace ) for the class to prevent name collisions with other classes of the same name (for example classes from third party libraries we want to use in our project) and to logically group classes that belong together.
```
lukasbestle, line 236: I'd explain here that the `Page` class becomes `Kirby\Cms\Page` because of the namespace. That makes the following section easier to understand.

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+```
+
+### Inheritance (`extends`)
+
+If a class extends another class, we speak of inheritance, one of the big advantages of OOP. In this case, the `Page` class is a child class of the `ModelWithContent` class and inherits all its methods and properties.
+
+However, it doesn't just inherit them but can modify existing (public or protected) methods or add additional properties and methods. Basically, the `ModelWithContent` class can be regarded as the more generic class, and `Page` is a more specific child class of this class. Other Kirby classes that inherit from `ModelWithContent` are `Site`, `File` and `User`.
+
+If we return to the `Book` example from above, it would probably make sense to extend the basic `Book` class with more specific child classes, for example for ebooks that are different from paper books. A paper book, for example, has pages you can count, an ebook doesn't. A paper book probably has a hard- or softcover, whereas an eBook has different file formats like PDF or Mobi etc. Inheritance is therefore a powerful way to modularize code while at the same time keeping code dry. You only have to redefine methods or extend where necessary.
```
lukasbestle:
```suggestion
If we return to the `Book` example from above, it would probably make sense to extend the basic `Book` class with more specific child classes, for example for eBooks that are different from paper books. A paper book, for example, has pages you can count, an eBook doesn't. A paper book probably has a hard- or softcover, whereas an eBook has different file formats like PDF or Mobi etc. Inheritance is therefore a powerful way to modularize code while at the same time keeping code DRY ("don't repeat yourself"). You only have to redefine methods or extend where necessary.
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+<info>
+In the same way, `Pages`, `Users`, `Files`, `Roles`, `Languages`, `Fields` etc. all extend the `Collection` class which is why these classes basically share the same methods.
```
lukasbestle:
```suggestion
In the same way, `Pages`, `Users`, `Files`, `Roles`, `Languages`, `Fields` etc. all extend the `Collection` class, which is why these classes basically share the same methods.
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+And you will of course find a few more classes that inherit from other classes.
+</info>
+
+When you create a (link: docs/guide/templates/page-models text: page model), you create a child class of the `Page` class that inherits all it's properties and methods and thus allows you to extend this class with your own methods and properties, or to redifine existing methods in the same way that the `Page` class extends the `ModelWithContent` class.
+
+## Traits
+
+With the `Page` class we started with one that comes at you like a sly old dog. Lot's of stuff that you wouldn't have expected.
```
lukasbestle:
```suggestion
With the `Page` class we started with one that comes at you like a sly old dog. Lots of stuff that you wouldn't have expected.
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,468 @@
...
+
+```php
+$p = page('notes');
+dump($p->title());
+$p = $p->update([
+    'title' => 'Blog'
+]);
+dump($p->title());
+```
+Now the changes will be reflected without reloading the page, because we override the original `$p` variable and assign the new page object to it.
```
lukasbestle:
```suggestion
Now the changes will be reflected without reloading the page, because we overwrite the original `$p` variable and assign the new page object to it.
```

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
texnixe: :laughing:

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
texnixe: Great idea, thanks! Getting too abstract sometimes...
I think I'd rather stay with the book analogy though, to keep it consistent, instead of introducing apples ;)

texnixe:
Thanks a lot @lukasbestle for taking the time to go through this long recipe and for your very valuable feedback.

content/1_docs/2_cookbook/9_setup/0_understanding-oop/recipe.txt
lukasbestle: You are right, otherwise it really is an apple to books comparison. 😆

PR #1066, opened 2020-11-18
Fix redundant sortBy in cheatsheet-article.txt

renestalder:
Remove the redundant `sortBy` property from the `sections/events.yml` example
and update the first occasion to use desc sorting, assuming it is meant to show
the most recent events first


texnixe:
Thanks!

PR #1074, opened 2020-12-01
Update : <k-input> to <k-text-input>

kevinvennitti:


bastianallgeier:
You should always use the wrapping k-input component with type="text" instead of the low-level k-text-input component.

PR #1082, opened 2020-12-07
Reference refactor

distantnative:
**First merge https://github.com/getkirby/getkirby.com/pull/1150**

To make this PR somehow manageable, I have excluded the content file renames from it. To rename all content files, run `./scripts/reference-rename`. Some content files will need smaller changes afterwards (e.g. the objects need a `title` field for their typical notation, e.g. `$page` ). I omitted these too, to avoid merge conflicts later on. We can finish those on the `develop` branch after merging this PR.

So what does this PR do?
- Refactors all page models for reference pages
- To better manage the refactoring projects, this PR moves some logic stuff from the big old `site` plugin to a dedicated `reference` plugin. This can be temporary - once we have refactored the `site` plugin itself, this could be merged into a single plugin again.
- The plugin introduces two page models as foundations: `ReflectionPage` and `ReflectionPage`. The reference page models all extend one of these.
- All reference page models/templates have been renamed with a `reference.` prefix to better organize them
- The actual reference page models live now in `site/models`
- Templates and snippets for the reference pages have been refactored and adapted, e.g. all reference snippets now reside in `site/snippets/reference` (instead of using `cheatsheet.` prefixes)

There is plenty more room for further improvements to the models. I would however, merge this first as a foundation for future features (e.g. proxy classes...)


lukasbestle:
@distantnative What is the plan forward for this? Are you still working on it or can this be reviewed when we find the time after the 3.5 release? I think it makes sense to get this merged soon to avoid that you have to do it yet another time.

distantnative:
My plan would be to get all the 3.5 release stuff out of the way, then see if this PR still works. If needed, make changes/fixes - but then, yes, merge as fast as possible before we get into other website things.

distantnative:
Some content fields have to be changed as well to match the new logic better. But that really only makes sense right before merging, so we don't run into merge conflicts. So I am waiting until all the 3.5 stuff is through.

distantnative:
Maybe the site plugin refactor should go first.

texnixe:
I'll leave my comments here as I come across them...

- Loading aliases seems to be slow (last nav entry in the sidebar)
- Not related to the PR in particular: But the information below the reference overview pages is easily overlooked, maybe we can find a better way to make these noticeable. People often overlook that they can scroll down to get more information, e.g. on https://getkirby.com/docs/reference/objects/page
- Edit on GitHub links don't work anymore, probably due to the renaming...
- Apart from that and the issues with PHP 8, I didn't run into any other issues…

distantnative:
@texnixe 
Thanks for testing!
I'll take a look at No. 1 and 3 this evening.
No. 2 I totally agree. In an earlier version I had changed it in some places but not in a good way. I'd put it as a separate package/task to not blow this PR up any further. But definitely something we should tackle - maybe when we look at refactoring the Reference template/design

distantnative:
1. Should be solved by adding caching for reference lookups. Separate project. Added to our Notion list.
2. Also added as project to Notion list.
3. Yes, will work again once file are merged onto `master`
4. 🙌

PR #1086, opened 2020-12-11
Update cheatsheet-article.txt

S1SYPHOS:


distantnative:
I don't think this would work (toggle field with options)

afbora:
It works but without reading `options` :)) doesn't need anyway. 
In order not to get confused, `options` should be removed.

S1SYPHOS:
Changed it already .. waiting for it to show up ...

S1SYPHOS:
Well, now there are two of them - never should have tried GitHub direct GUI editing ..

distantnative:
Is this actually about an inaccuracy or just a personal preference to rather use `toggle` instead of `radio`?

S1SYPHOS:
From my PoV it's about consistency.

distantnative:
@S1SYPHOS consistency in what sense? I'm not sure I understand

S1SYPHOS:
@distantnative Well, looking at it now, it might as well be a radio, but ..

1. the other "switch" in the article is a toggle, too (consistency)
2. semantically, a toggle (between **only two** states) makes more sense imho

texnixe:
Personally, I think these are different examples altogether, so consistency is probably not necessary here. It's always good to show different options.

A toggle is sometimes harder to understand for some people, who'd probably prefer to use a radio field or a checkbox for only two options. So I don't think that the only option for choosing between two options should necessarily be a toggle field.

distantnative:
I agree, let's stick with the radio field for this example.

PR #1087, opened 2020-12-11
Update cheatsheet-article.txt

S1SYPHOS:


afbora:
I guess that was also wrong. It should have been `toggle` on this 😅

texnixe:
That's because of the two PRs now, I think you better close them and do them in one request.

PR #1092, opened 2020-12-13
Fixing assumed mistake in guide

sebastiangreger:
If I am not mistaken, the described action makes the former draft an `unlisted` page, not `listed`. Only after adding a sorting number, it becomes `listed`.


texnixe:
Yes, of course, thank you!

PR #1115, opened 2020-12-19
Add AlpineJS

distantnative:
As a first step to a build-less setup and modernising the whole website, this PR currently

- Adds AlpineJS loaded from `cdn.jsdelivr.net` to every page
- Removes the old `try` template JS from the build process and replaces it with a tiny alpineJS implementation

**Open questions:**
- [ ] Load from CDN or self-hosted? GDPR-wise self-hosted might be better/requried.
- [ ] Load alpineJS on every page or define on which pages?


lukasbestle:
@distantnative I'd say we shouldn't use external CDNs for the mentioned reasons. However I cannot say anything about the question whether to keep the library in general.

PR #1131, opened 2020-12-30
Redirects: easier to manage

distantnative:
A little plugin that allows us to register redirects simpler in `site/config/redirects.php`.
Also those only get checked if no actual page exists, so taking a bit load off the main router.

This will help us a lot to easily set up redirects when we want to reorder the site a bit, e.g. the reference structure, cookbook categories...

### DIFF
--- a/site/plugins/redirects/index.php
+++ b/site/plugins/redirects/index.php
@@ -1,0 +1,57 @@
+<?php
+
+use Kirby\Http\Router;
+
+/**
+ * Plugin creates redirect routes (using the `go()` helper) that only 
+ * take over if no actual page/route has been matched.
+ * 
+ * The rredirects get defined in `site/config/redirects.php` in an array 
+ * with the old pattern as key and the target page/URL as value. Placeholders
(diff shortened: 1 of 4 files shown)


site/plugins/redirects/index.php
```diff
@@ -0,0 +1,42 @@
...
+
+                        // turn into routes array
+                        $routes = array_map(function($from, $to) {
+                            return [
+                                'pattern' => $from,
+                                'action'  => function (...$parameters) use ($to) {
+
+                                    // resolve callback
+                                    if (is_callable($to) === true) {
+                                        $to = call_user_func($to, ...$parameters);
```
johannschopplich: @distantnative Your plugins are always imaginative & inventive. I learn a lot reading the code. Thank you!

Isn't `$to(...$parameters)` more PHP7-esque than using `call_user_func()`?

distantnative: Thank you, @johannschopplich!

bastianallgeier (approved):

PR #1141, opened 2021-01-03
Update migration recipe: route arguments

distantnative:


content/1_docs/2_cookbook/9_setup/0_migrate-site/recipe.txt
```diff
@@ -272,6 +272,10 @@ return function ($kirby) {
 };
 ```
 
+<since v="3.1.3">
+In your route, you can also pass all arguments explicitely to the `->render()` method to make them directly accedible in the controller. (link: docs/guide/templates/controllers#arguments-from-page-render-in-route text: Read more ›)
```
texnixe: Should be `accessible` instead of `accedible`

bastianallgeier (approved):

PR #1143, opened 2021-01-04
Add charset UTF-8 to locale string

m-podolski:
On my machine (Ubuntu, NginX, PHP 7.4) the strftime-function only translates successfully when the ".utf-8" extension is present in the locale string.

### DIFF
--- a/content/1_docs/1_guide/7_languages/1_introduction/guide.txt
+++ b/content/1_docs/1_guide/7_languages/1_introduction/guide.txt
@@ -57,7 +57,7 @@ return [
   'code' => 'en',
   'default' => true,
   'direction' => 'ltr',
-  'locale' => 'en_US',
+  'locale' => 'en_US.utf-8',
   'name' => 'English',
 ];
 ```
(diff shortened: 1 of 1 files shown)


texnixe:
Hm, what to put there really depends on your installed locales, so I wonder if adding `utf-8` in general is a good idea?

lukasbestle (commented):
Unfortunately this depends on the operating system and setup. Some operating systems (like macOS) use `.UTF-8` or `.UTF8` instead. So I agree with @texnixe that we shouldn't change it in the example. But we could add a note about this issue with the tip to try the different variants if it doesn't work.

distantnative:
What would be a good phrasing? I'm happy to add the info box quickly, but not 100% sure what it should say...

"Depending on your operating system and setup, it might be necessary to add a suffix to your local code that specifies the encoding format. This ensures that functions , e.g. `strftime` can function properly. If you are encountering issue with it, try adding e.g. `.utf-8` to your locale codes. Some operating systems however use different notations such as `.UTF-8` or `.UTF8`"

texnixe:
If you encounter issues with functions like `strftime` that depend on the locale setting, you might have to add a suffix to the local code that specifies the encoding format. Depending on your operating system and setup, try adding e.g. .utf-8 to your locale codes. Some operating systems however use different notations such as .UTF-8 or .UTF8.

You can list all available locales on your system by typing `locale -a` on the command line.

distantnative:
Added manually.

PR #1146, opened 2021-01-05
Reference lookup also on local

distantnative:
I don't know why we ever deactivated it, but it seems to work without issues also on local. I would suggest to streamline and remove that option.

### DIFF
--- a/site/config/config.getkirby.com.php
+++ b/site/config/config.getkirby.com.php
@@ -17,6 +17,5 @@ return [
     ],
     'keycdn' => [
         'domain' => 'https://assets.getkirby.com',
-    ],
-    'referenceLookup' => true
+    ]
 ];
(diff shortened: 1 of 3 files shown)


bastianallgeier:
I deactivated it because some pages load forever with the reference lookup. Eventually, we should introduce caching for this.

distantnative:
Do you have an example? I tried to find one (cause I thought that was the reason) but couldn't spot one. However, caching indeed would be great. Maybe once the reference and site plugin is refactored, that would fit in nicely

bastianallgeier:
@distantnative when I was working on the release page for 3.5 I had some really terrible performance problems.

distantnative:
Ok, let's close this. And only remove it in combination with adding caching (which shouldn't be too hard) once the refactors have been merged.

PR #1148, opened 2021-01-05
Add Ahmet to team page

distantnative:


afbora:
Just a suggestion: It would be better if the links in team and contributors are opened in new tab with `rel=nofollow`?

distantnative:
I think in terms of accessibility it's always better to let the user choose if to open in same tab or new tab.

afbora (approved):
❤️

bastianallgeier:
@afbora so good to finally have you on the team page!

PR #1150, opened 2021-01-06
Refactor `site` plugin

distantnative:
## What it does
- Integrates all plugins into new `site` plugin
- Migration of many improvements (e.g. Marsdown) from tailwind project
- Plugin files from the reference refactoring (https://github.com/getkirby/getkirby.com/pull/1082) have already been included on this PR, but not any of the other changes from the reference refactoring (meant to make this PR easier to read, but we'll need both PRs for the site to properly function)

## What needs especially attention
- There might be still some styling issues (due to Marsdown introduction). But it looked to me as if we could start merging this to `develop` where we can work on the fine-tuning (or go straight to refactoring of templates and snippets).

## To be discussed
 - Integrating all separate plugins into one  makes our setup a lot easier to understand and maintain for us - at the same time the separate plugins aren't as easily to copy by others as they're all integrated into one


site/plugins/site/composer.json
```diff
@@ -1,11 +1,15 @@
...
-    "require": {
-        "phpdocumentor/reflection-docblock": "^3.1"
+  "name": "getkirby/site",
+  "autoload": {
+    "psr-4": {
+      "Kirby\\": "src/"
     }
+  },
+  "require": {
+    "algolia/algoliasearch-client-php": "^1.25",
```
afbora, line 9: Can we switch to v2? Because v1 doesn't support PHP8.

afbora:
`/site/plugins/site/src/Algolia/vendor` directory is unnecessary.

site/plugins/site/extensions/tags.php
```diff
@@ -0,0 +1,232 @@
...
+            'comment'  => $comment,
+        ];
+
+    }
+
+    ksort($table);
+
+    return $table;
+
+}
```
afbora, line 78: What if we move to `helpers`?

site/plugins/site/extensions/tags.php
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * Fetch prop definitions from Fields and Sections
+ * and create an options table for it.
+ *
+ * @param array $definition
+ * @return array
+ * @todo refactor/deprecate
+ */
+function componentOptions(array $props) {
```
afbora: What if we move to `helpers`?

site/plugins/site/extensions/tags.php
```diff
@@ -0,0 +1,232 @@
...
+            $default = $parameter->getDefaultValue();
+        } catch (Exception $e) {
+            $default = null;
+        }
+
+        if ($docComment = $reflection->getDocComment()) {
+            try {
+                $docBlock = new DocBlock($docComment);
+                $comment  = trim($docBlock->getSummary());
+                $comment  = str_replace(PHP_EOL, ' ', $excerpt);
```
afbora:
```suggestion
                $comment  = str_replace(PHP_EOL, ' ', $comment);
```

site/plugins/site/src/Compressor/Cache.php
```diff
@@ -30,11 +32,4 @@ public function set(string $key, $value, int $minutes = 0): bool
         parent::set($key, $value, $minutes);
```
afbora:
```suggestion
        return parent::set($key, $value, $minutes);
```
Nothing was returning.

site/plugins/site/src/Marsdown/Marsdown.php
```diff
@@ -0,0 +1,350 @@
...
+
+        $infostring = explode(' ', $infostring, 2);
+        $language   = $infostring[0];
+
+        if (sizeof($infostring) === 2) {
+            // Block with caption
+            $caption  = $infostring[1];
+            $openChar = $caption[0];
+
+            if ($openChar === '"' || $openChar === '"') {
```
afbora: Are these two the same or am I double seeing? 🙂

site/plugins/site/src/Marsdown/Marsdown.php
```diff
@@ -0,0 +1,350 @@
...
+                $captionLength = strlen($caption);
+                $lastChar      = $caption[$captionLenght - 1];
```
afbora:
```suggestion
                $lastChar      = $caption[$captionLength - 1];
```

site/plugins/site/extensions/fieldMethods.php
```diff
@@ -0,0 +1,30 @@
...
+            '$\`(.*?)\`$',
+            '<code>$1</code>',
+            $field->value
+        );
+        return $field;
+    },
+
+    /* Legacy */
+    'replace' => function ($field, $replace) {
+        $field->value = Str::template($field->value(), $replace);
```
afbora, line 23: We can import all the classes we use in the plugin instead of using alias?

site/plugins/site/extensions/components.php
```diff
@@ -0,0 +1,72 @@
+<?php
+
+use Kirby\Cachebuster\Cachebuster;
+use Kirby\Cms\App;
+use Kirby\Cms\FileModifications;
```
afbora: There are several unused imports

site/plugins/site/extensions/routes.php
```diff
@@ -0,0 +1,41 @@
...
+    [
+        'pattern' => 'sitemap.xml',
+        'action' => function () {
+            return SiteMeta::sitemap();
+        }
+    ],
+    [
+        'pattern' => 'open-search.xml',
+        'action' => function () {
+            return SiteMeta::search();
```
afbora, line 23: I couldn't review the following file.
Related file: `site/plugins/site/src/Meta/snippets/search.php`

Shouldn't this be added to the beginning of the base64 code? `data:image/png;base64,XXX`

Like that
````xml
<Image width="64" height="64" type="image/png">data:image/png;base64,<?= (new Kirby\Toolkit\File(kirby()->root('index') . '/opensearch.png'))->base64() ?></Image>
````

Document:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs

afbora:
Wouldn't it be great if the site plugin had unit tests?  What do you think?

site/plugins/site/composer.json
distantnative: I only migrated the existing plugin into a new structure. I would see upgrading (and making changes so the plugin still works) as a separate little project after merging this.

site/plugins/site/extensions/components.php
distantnative: Several? For me only `FileModifications ` in this.

site/plugins/site/extensions/fieldMethods.php
distantnative: I am not sure what you mean here 😬

site/plugins/site/extensions/routes.php
distantnative: I know too little of XML to say whether it's needed there or not. I find different examples online. Migrated it as it is used currently on the website (without).

distantnative:
@afbora As with many things, unit tests would be great :D But I wouldn't blow up this PR any more. Could be it's own project after that to write unit tests. But there's also so much more to do :D

site/plugins/site/extensions/fieldMethods.php
afbora: 😅 I mean it's not a bug, it's a standard. Let's import the classes we use alias.

site/plugins/site/extensions/components.php
afbora: I can see also in following classes:
- site/plugins/site/src/Algolia/Search.php
- site/plugins/site/src/CodeBlock/KirbyContent.php

site/plugins/site/extensions/routes.php
afbora: Then let's leave it as it is 👍

afbora:
@distantnative You're right, then let's do the unit tests in the next step 👍

site/plugins/site/extensions/fieldMethods.php
distantnative: Now understood! :)

afbora: @distantnative There are uses that need to be imported in other files.

bastianallgeier (commented):
I think the PR is great and pretty much ready to be merged to develop. I found an issue with the SVG icons in the Cookbook and Plugin templates. That's probably because they are using strokes and not fills. But that's something we can still tweak and maybe even replace the icons. 

I also found an old compressor plugin. I think we actually never use it anywhere. We should check that and remove it. 

Otherwise I really love that we have everything in a single plugin now.

distantnative:
Removed the compressor plugin. Indeed unused.

I agree, there might be smaller styling inconsistencies. But those will be changed soon anyways.
Still unsure whether this should go to `master` or if this is rather a first block of the whole refactor...

It's just really tricky as we constantly keep adding content changes to `master`

bastianallgeier:
I'm torn as well. It already seems pretty stable to me and I wouldn't mind causing a few minor glitches on the old site. It could actually help us fix them in a real-world test field before we even move it over to the redesign repo.

bastianallgeier:
@distantnative could you take care of rebasing the develop branch and fixing the conflicts?

distantnative:
@bastianallgeier done

PR #1151, opened 2021-01-06
doc: details on creating translated pages via API

pReya:
I learned this today on the forum: https://forum.getkirby.com/t/x-language-header-in-rest-api-calls-is-ignored/

Should definitely be part of the documentation in my opinion.

Let me know if this needs more polishing or any other corrections you might have.


texnixe:
Thanks for adding this information.

PR #1152, opened 2021-01-07
Add hashed assets plugin

johannschopplich:
Use hashed assets with the `css()` and `js()` helpers without changing the asset paths passed.

https://github.com/johannschopplich/kirby-hashed-assets


lukasbestle:
Thanks! 👍

PR #1153, opened 2021-01-07
Add locked pages plugin

johannschopplich:
Protect pages that you want to hide from unwanted views with a password.

https://github.com/johannschopplich/kirby-locked-pages


lukasbestle:
Thanks! 👍

johannschopplich:
Good catch. Will remember to uppercase "Panel". 🙂

PR #1158, opened 2021-01-09
Integrate CSS framework

distantnative:
@bastianallgeier this could be one way how we integrate the CSS framework into getirby.com by keeping it in its own repository. It does feel a bit cumbersome to do changes in a whole separate repo while working on it though. On the other hand, it would offer us the chance to keep the framework uncoupled from one specific integration (getirby.com, kits, panel).

I put the `legacy.css` last to avoid breaking changes in the styling for now.
Moving forward we could rather import the `legacy.css` in each template-specific file. So it is still present for the unrefactored ones - but every template we have refactored can live without it


lukasbestle:
I think it makes sense to keep the files separate, but with a submodule we will need to update it all the time. What we could do is to create a workflow in the `css` repo that automatically commits a submodule update to all dependent repos whenever there are changes in the `css` repo.

Actually we would then not even need the submodule, which would make setting up the repos easier (especially the kits and Kirby itself).

distantnative:
Sounds good!
It still doesn't take away the "pain" of working in two repos. That's a tradeoff we just have to decide.

PR #1159, opened 2021-01-09
Assets workflow: run on PR approval

distantnative:


.github/workflows/assets.yml
```diff
@@ -1,26 +1,36 @@
...
 on:
-  push:
-    paths:
-      - 'www/assets/**'
-  pull_request:
+  pull_request_review:
+    types:
+      - submitted
     paths:
       - 'www/assets/**'
```
lukasbestle: If I read the docs correctly, the `pull_request_review` event does not support the `paths` option, so the workflow will run on all approved PRs. I don't know if there is a way around that.

.github/workflows/assets.yml
```diff
@@ -1,26 +1,36 @@
...
 
 jobs:
-  assets:
+  build:
+    if: github.event.review.state == 'approved'
     runs-on: ubuntu-16.04
```
lukasbestle: Does the workflow use the old Ubuntu version on purpose?

.github/workflows/assets.yml
```diff
@@ -1,26 +1,36 @@
...
+      - name: Checkout getkirby.com
+        uses: actions/checkout@v2
         with:
           ref: ${{ github.head_ref }}
-      - uses: actions/setup-node@v2
+
+      - name: Setup node environment
+        uses: actions/setup-node@v2
         with:
           node-version: 12
```
lukasbestle: Do we explicitly need Node 12? We would need to update this manually when Node 12 is no longer supported. If no `node-version` is given, GitHub Actions defaults to one anyway.

.github/workflows/assets.yml
```diff
@@ -1,26 +1,36 @@
...
-      - run: npm install
-      - run: npm run build
-      - uses: stefanzweifel/git-auto-commit-action@v4
+
+      - name: Install npm
+        run: npm install
```
lukasbestle: I'd recommend using `npm ci` instead, which can use the global cache in a better way. See the workflow in the Kirby repo.

.github/workflows/assets.yml
lukasbestle: I thought about this again and think it's fine that the workflow always runs on all PRs: If the build script only changes the dist files if the src files changed (i.e. if the dist files stay the same if the src is unchanged), it's a non-issue to run the workflow regardless as it won't result in a commit.

distantnative: You are right, it's not really an issue. Just annoys me that it will run even for PRs the only change a content file XD

.github/workflows/assets.yml
distantnative: No,, I think that's just a copy-and-paste artifact of me looking at some guides.

.github/workflows/assets.yml
distantnative: Having encountered issues with new node versions in the past, I thought it might be useful to pin the version and purposefully change update it. But we can try with default as well.

.github/workflows/assets.yml
lukasbestle: I'm curious: Why does it annoy you?

.github/workflows/assets.yml
lukasbestle: `ubuntu-latest` is recommended.

.github/workflows/assets.yml
distantnative: Cause it's running a check that's totally unnecessary for the PR then. And it delays showing the status of the Pr as the test hast to be finished first. It's a non-issue as you say, but it bothers me especially because it would be so easy to define the rule for when it should run and when not. But I'll be able to live with it ;)

lukasbestle (approved):

lukasbestle (approved):

PR #1161, opened 2021-01-09
Add recipe how to update blocks programmatically

texnixe:


content/1_docs/2_cookbook/3_templating/0_update-blocks-programmatically/recipe.txt
```diff
@@ -0,0 +1,67 @@
...
+    'text'  => 'This is a heading',
+    'level' => 'h2',
+  ],
+  'type'     => 'heading',
+]);
+
+$quote = new Kirby\Cms\Block( [
+  'content' => [
+    'text'     => 'Time flies like an arrow; fruit flies like a banana',
+    'citation' => 'a href=\"https://en.wikipedia.org/wiki/Anthony_Oettinger\" rel=\"noopener noreferrer nofollow\">Anthony Oettinger</a>',
```
afbora: Missing start `<a` tag.

texnixe: Thanks!

afbora (approved):

lukasbestle (commented):
To be honest I think this is another example of a missing core feature. Manually encoding the data as JSON doesn't feel right...

afbora:
Yes, I absolutely agree, the `->toJson()` method like `->toArray()` can be added to the block class.

lukasbestle:
I don't think that's the right way. I was thinking more about a method to update a `blocks` field directly. Same for `structure` actually. This way everything will still work if we change the way these fields store their data.

texnixe:
Great idea, structures have always been a pain to update all these years. Problem is, as always, that people need a solution now, not in x months, years ...

afbora:
Ahh, wait a minute. This way already supports both array and objects 😅 

````php
// array
$page->update([
  'text' => $blocks->toArray()
]);

// object
$page->update([
  'text' => $blocks
]);
````

https://github.com/getkirby/kirby/blob/3.5.0/src/Form/Field/BlocksField.php#L179-L183

lukasbestle (approved):

PR #1164, opened 2021-01-11
Add OOP recipe

texnixe:
Rewritten version...

### DIFF
--- a/content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
+++ b/content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
@@ -1,0 +1,462 @@
+Title: A brief intro to object oriented programming in PHP
+
+----
+
+Published: 2021-01-11
+
+----
+
+Description: Understanding classes, objects, member functions and common pitfalls
+
+----
+
+Text:
+
+## Intro
+
+When you have already tinkered with Kirby for a while, you are likely to have come across terms like classes, objects or methods. You've probably run into errors like "calling a member function x on null" and have then been asked to "check for an object before you call a method". Or have tried to use `$kirby` or `$site` in a plugin and been told to use `kirby()` and `site()` in this context instead. And then you've followed the advice but never really understood why.
+
+If that is true, then this recipe is for you. We will look into the basics of object oriented programming (OOP) with the goal of making you understand what this is all about.
+
+The Kirby core is written in PHP, a scripting language that still powers a majority of websites both in the backend and in the frontend.
+
+Like basically all professional PHP applications today, Kirby is based on an OOP (Object Oriented Programming) architecture which has many advantages over the procedural approach you often find in older PHP programs. The idea behind this architecture is to encapsulate data and the methods that work with this data, in order to produce modular, extendable and maintainable code.
+
+<warning>
+No images, lots of text😉. But at least some code to play with. So get some coffee or tea.
+</warning>
+
+## Prerequisites
+
+If you want to code along with this recipe, you need the following:
+
+- A web server (ideally on your (link: docs/cookbook/setup/development-environment text: local machine))
+- A current version of PHP
+- A Kirby installation , either (link: docs/guide/quickstart text: Starterkit or Plainkit)
+- A code/text editor of your choice
+
+## What is OOP?
+
+Object Oriented Programming (OOP) is all about objects, their properties and their methods.
+
+Let's consider a real-world object. Such an object could be a book in your library (or a computer, a smartphone, a piece of furniture…). It has certain properties like an author, a title, an ISBN number, a price tag and a cover, and it was probably published by a publishing house. Its methods would be the actions you could do with the book, e.g. you can read it, you can give it away as a present, put it back in your shelf or throw it away if you don't like it. And while all the books in your library are different, they nevertheless share the same general properties and methods.
+
+OOP uses the same paradigms for data structures in a program. The properties and methods of an object like the book are described in a class, which acts as the model or blueprint for every book you might have in your application, whereas an object is an instance of this class, i.e. a specific embodiment of the class (the specific book in your library).
+
+In a larger software program, you typically have multiple classes whose objects can interact with each other (like a book, a bookshop, the customers that buy the books in your shop, invoices for sold books etc).
+
+### A simple class example
+
+Time for the first coffee. Let's start with a simple example and do some coding to see how we can actually create such a class in PHP and how we can create (instantiate) objects from that class. We define a class called `Book` with some properties that should apply to all objects of this class, and methods to work with the data of these objects.
+
+In your webserver's (glossary: document-root text: document root), create a new folder `book` and inside of this folder a new file called `book.php` with the class definition code from above. Copy the code below into this file.
+
+```php "/book/book.php"
+<?php
+class Book
+{
+    protected $author;
+    protected $title;
+    protected $isbn;
+    protected $price;
+    protected $publisher;
+    protected $type;
+
+
+    public function __construct(string $title, string $author)
+    {
+        $this->title  = $title;
+        $this->author = $author;
+    }
+
+    public function getTitle(): string {
+        return $this->title;
+    }
+
+    public function getAuthor(): string {
+        return $this->author;
+    }
+
+    public function getPrice(): float {
+        return $this->price;
+    }
+
+    public function getPriceWithCurrencySymbol(string $symbol): float {
+        return $symbol . $this->price;
+    }
+
+    public function getPublisher(): string {
+        return $this->publisher;
+    }
+    
+    public function getIsbn(): string {
+        return $this->isbn;
+    }
+      
+    public function getType(): string {
+        return $this->type;
+    }
+
+    public function setTitle(string $title): void {
+        $this->title = $title;;
+    }
+
+    public function setAuthor(string $author): void
+    {
+        $this->author = $author;
+    }
+
+    public function setPrice(float $price): void {
+        $this->price = $price;
+    }
+
+    public function setPublisher(string $publisher): void {
+        $this->publisher = $publisher;
+    }
+    
+    public function setIsbn(string $isbn): void {
+        $this->isbn = $isbn;
+    }
+
+    public function setType(string $type): void {
+        $this->type = $type;
+    }
+    
+    public function isEbook(): bool {
+        return $this->type === 'ebook';
+    }
+}
+```
+If you find books boring, create a similar class for games, films, smartphones or whatever you prefer.
+
+Now for some explanations…
+
+## Describing a class
+
+For starters, we define our new `Book` class with the keyword `class`. Inside the class, we define the properties that objects of this class should have, like a title, an author, a publisher etc. Of course, a book could have many more properties, and if you feel like it, you can extend this example yourself.
+
+### Properties and methods
+
+#### Visibility
+
+Class properties (variables and constants) and methods can have different (link: https://www.php.net/manual/en/language.oop5.visibility.php text: visibility statuses): private, protected or public. 
+
+We define a protected property with `protected` followed by the variable for the property, e.g. `protected $author`. The properties of a class usually should not be public to prevent users from assigning unexpected values.
+
+Public properties and methods like the setters and getters in our `Book` class can be called anywhere, not only inside the class where they are defined, but also inside other classes or scripts when we work with objects of that class.
+
+Protected properties and methods can be called within the defining class itself or within the children of that class, but not outside of these classes.
+
+Private properties and methods are even more restricted, they can only be used inside the defining class itself, not anywhere else.
+
+#### Getters and setters
+
+Because the properties are not public, we cannot access them directly, but have to create so called *getters* that return the value of a property (`getPrice()`, `getTitle()` etc.) and *setters*, which assign a value to the property of an object (`setPrice()`, `setTitle()` etc.). Getters and setters are often prefixed with `get-`/`set-` to indicate their purpose, but this is not required and you can name them what you want (although it is good practice to give methods, variables etc. speaking names and not name them foo or bar);
+
+#### Other methods
+
+Methods do stuff with the data of an object. In addition to the setters and getters we can therefore have all sorts of other methods, e.g. methods that check if something is true about the object (`isEbook()`), methods to modify the data in some way, e.g. return the price prefixed with a currency symbol (`getPriceWithCurrencySymbol()`) or anything else.
+
+<info>
+Not every method in a class must be public. Methods that are only relevant inside the class or its decendents should therefore also be protected.
+</info>
+
+### $this
+
+You may have noticed that we use `$this` in all setters and getters above. `$this` is a pseudo-variable in PHP OOP that refers to the calling object, i.e. an instance of the class to which the method belongs.
+
+### Trying it out
+
+Enough talk, let's do some more coding: We will be creating new objects, set their properties and retrieve them.
+
+Next to the `book.php` file from above, create an `index.php` file with the following content:
+
+```php "/book/index.php"
+<?php
+// make PHP display errors on screen
+ini_set('display_errors', 1);
+// require the book class
+require_once __DIR__ . '/book.php';
+```
+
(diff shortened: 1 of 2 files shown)


lukasbestle:
@texnixe What has changed compared to the first PR #1064?

texnixe:
@lukasbestle I have removed some parts like the stuff with the page class in favor of more actual coding and some more advanced concepts, because it thought it was too long-winding before. So it's basically a new start while I have of course also kept parts of it.

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+
+Text:
+
+## Intro
+
+When you have already tinkered with Kirby for a while, you are likely to have come across terms like classes, objects or methods. You've probably run into errors like "calling a member function x on null" and have then been asked to "check for an object before you call a method". Or have tried to use `$kirby` or `$site` in a plugin and been told to use `kirby()` and `site()` in this context instead. And then you've followed the advice but never really understood why.
+
+If that is true, then this recipe is for you. We will look into the basics of object oriented programming (OOP) with the goal of making you understand what this is all about.
+
+The Kirby core is written in PHP, a scripting language that still powers a majority of websites both in the backend and in the frontend.
```
lukasbestle: In the frontend?

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+
+Like basically all professional PHP applications today, Kirby is based on an OOP (Object Oriented Programming) architecture which has many advantages over the procedural approach you often find in older PHP programs. The idea behind this architecture is to encapsulate data and the methods that work with this data, in order to produce modular, extendable and maintainable code.
```
lukasbestle:
```suggestion
Like basically all professional PHP applications today, Kirby is based on an OOP (Object Oriented Programming) architecture, which has many advantages over the procedural approach you often find in older PHP programs. The idea behind the OOP architecture is to encapsulate data and the methods that work with this data, in order to produce modular, extendable and maintainable code.
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+- A web server (ideally on your (link: docs/cookbook/setup/development-environment text: local machine))
+- A current version of PHP
+- A Kirby installation , either (link: docs/guide/quickstart text: Starterkit or Plainkit)
+- A code/text editor of your choice
+
+## What is OOP?
+
+Object Oriented Programming (OOP) is all about objects, their properties and their methods.
+
+Let's consider a real-world object. Such an object could be a book in your library (or a computer, a smartphone, a piece of furniture…). It has certain properties like an author, a title, an ISBN number, a price tag and a cover, and it was probably published by a publishing house. Its methods would be the actions you could do with the book, e.g. you can read it, you can give it away as a present, put it back in your shelf or throw it away if you don't like it. And while all the books in your library are different, they nevertheless share the same general properties and methods.
```
lukasbestle: I really like the more detailed examples for the book object! I'd leave out the other objects though as they are confusing before the book explanations. Alternative: Put the other objects at the end like this:

```suggestion
Let's consider a real-world object. Such an object could be a book in your library. It has certain properties like an author, a title, an ISBN number, a price tag and a cover, and it was probably published by a publishing house. Its methods would be the actions you could do with the book, e.g. you can read it, you can give it away as a present, put it back in your shelf or throw it away if you don't like it. And while all the books in your library are different, they nevertheless share the same general properties and methods. Other objects could be a computer, a smartphone or a piece of furniture. All have their own properties and methods.
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+
+OOP uses the same paradigms for data structures in a program. The properties and methods of an object like the book are described in a class, which acts as the model or blueprint for every book you might have in your application, whereas an object is an instance of this class, i.e. a specific embodiment of the class (the specific book in your library).
+
+In a larger software program, you typically have multiple classes whose objects can interact with each other (like a book, a bookshop, the customers that buy the books in your shop, invoices for sold books etc).
+
+### A simple class example
+
+Time for the first coffee. Let's start with a simple example and do some coding to see how we can actually create such a class in PHP and how we can create (instantiate) objects from that class. We define a class called `Book` with some properties that should apply to all objects of this class, and methods to work with the data of these objects.
+
+In your webserver's (glossary: document-root text: document root), create a new folder `book` and inside of this folder a new file called `book.php` with the class definition code from above. Copy the code below into this file.
```
lukasbestle: I'd name the file `book/Book.php` according to PSR standards so readers can learn it the "correct" way from the start.

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+    protected $type;
+
+
+    public function __construct(string $title, string $author)
+    {
+        $this->title  = $title;
+        $this->author = $author;
+    }
+
+    public function getTitle(): string {
```
lukasbestle: All of these methods should look like this:

```suggestion
    public function getTitle(): string
    {
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+
+### Properties and methods
+
+#### Visibility
+
+Class properties (variables and constants) and methods can have different (link: https://www.php.net/manual/en/language.oop5.visibility.php text: visibility statuses): private, protected or public. 
+
+We define a protected property with `protected` followed by the variable for the property, e.g. `protected $author`. The properties of a class usually should not be public to prevent users from assigning unexpected values.
+
+Public properties and methods like the setters and getters in our `Book` class can be called anywhere, not only inside the class where they are defined, but also inside other classes or scripts when we work with objects of that class.
```
lukasbestle:
```suggestion
Public properties and methods like the setters and getters in our `Book` class can be called from anywhere, not only inside from the class where they are defined, but also from other classes or scripts when we work with objects of that class.
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+
+Protected properties and methods can be called within the defining class itself or within the children of that class, but not outside of these classes.
```
lukasbestle:
```suggestion
Protected properties and methods can be called from the defining class itself or from the children of that class, but not from outside of these classes.
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+
+Private properties and methods are even more restricted, they can only be used inside the defining class itself, not anywhere else.
+
+#### Getters and setters
+
+Because the properties are not public, we cannot access them directly, but have to create so called *getters* that return the value of a property (`getPrice()`, `getTitle()` etc.) and *setters*, which assign a value to the property of an object (`setPrice()`, `setTitle()` etc.). Getters and setters are often prefixed with `get-`/`set-` to indicate their purpose, but this is not required and you can name them what you want (although it is good practice to give methods, variables etc. speaking names and not name them foo or bar);
```
lukasbestle:
```suggestion
Because the properties are not public, we cannot access them directly, but have to create so called *getters* that return the value of a property (`getPrice()`, `getTitle()` etc.) and *setters*, which assign a value to the property of an object (`setPrice()`, `setTitle()` etc.). Getters and setters are often prefixed with `get-`/`set-` to indicate their purpose, but this is not required and you can name them what you want (although it is good practice to give methods, variables etc. speaking names and not name them foo or bar).
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+```
+In Kirby, the `url()` method is a method that you find in the `Page`/`File` classes (and some others) and needs an object of this class to exist for it to work.
+
+But `page('non-existent')` or `$page->image('non-existent.jpg')` would return `null` and not an object, just like in our last error in the book example.
+
+Let's create another error:
+
+```php "/book/index.php"
+// …
+$book2 = new Book('Beloved', 'Toni Morrison')
```
lukasbestle:
```suggestion
$book2 = new Book('Beloved', 'Toni Morrison');
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+
+If we think that eBooks are sufficiently different from printed books (and for the purposes of this recipe we do), we might come to the conclusion that they are better off in their own `Ebook` class. However, we do not really want to copy all the code we defined in the `Book` class for the properties both classes share. That's where inheritance comes into play.
+
+If a class extends another class, we speak of inheritance, one of the advantages of OOP. A child class inherits all methods and properties of the parent class, like a real-life child inherits certain "properties" from its parents.
+
+However, in the same way that a child is not an exact copy of one of its parents, a child class can modify the inherited properties and methods and add additional ones. Inheritance is therefore a way to diversify while at the same time keeping code DRY ("Don't Repeat Yourself"). You only have to redefine or extend where necessary.
+
+But enough theory, let's just try it out with a little help of more coffee, and actually extend the `Book` class. In a new file `ebook.php` in the same folder as above, create the `Ebook` child class and add a new property and its setter and getter.
```
lukasbestle: As above: `Ebook.php`

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+    //… code as before
+}
+```
+
+```php "/book/ebook.php"
+<?php
+
+namespace Cookbook\Books;
+
+class EBook {
```
lukasbestle: The class was named `Ebook` above with lowercase `b`. Both are fine, but IMO they should be consistent. This also has an influence on the file name.

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+$book  = new Book(…);
+$ebook = new EBook(…);
+//…
+```
+
+## Static properties and methods
+
+Static properties and methods can be called directly without creating an instance of a class using the `new` keyword.
+
+You have probably come around such static methods when using Kirby. Same classes in the Toolkit, for example the `A`, Str` or 'Html` classes, have only static methods.
```
lukasbestle:
```suggestion
You have probably come around such static methods when using Kirby. Some classes in the Toolkit, for example the `A`, Str` or `Html` classes, have only static methods.
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+
+That we don't have to explicitly instanciate an object is not the only difference, we also call these methods differently, for example:
+
+```php
+echo Str::upper('Kirby is my favorite CMS');
+```
+
+Instead of the arrow syntax used for methods that are called on an instance of a class like `$page->update()`, a double colon (or (link: https://en.wikipedia.org/wiki/Scope_resolution_operator text: "Paamayim Nekudotayim")) is used to access a static method or property of a class.
+
+Fetch another coffee or tea and then let's add a static method to our `Book` class as well, so that we can create the new object statically instead.
```
lukasbestle:
```suggestion
Grab another coffee or tea and then let's add a static method to our `Book` class as well, so that we can create the new object statically instead.
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+
+```php "/book/book.php"
+<?php
+class Ebook extends Book
```
lukasbestle: This happens in the `Book` class, not `Ebook`.

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+```php "/book/index.php"
+$title = (new Ebook('Form Design Patterns', 'Adam Silver'))->getTitle();
+```
+
+vs.
+
+```php "/book/index.php"
+$title = Ebook::create('Form Design Patterns', 'Adam Silver')->getTitle();
+```
+
```
lukasbestle, line 464: I'd add a note here that this pattern can be useful sometimes, but probably shouldn't be used all the time.

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,462 @@
...
+## Conclusion && resources
+
+There would be so much more to learn and explore, but hopefully this recipe has lowered the entry barrier a bit, so that you are prepared to dive deeper yourself. If you explore the Kirby core classes, you will find a quite a few other OOP related terms like traits, abstract methods and classes, classes that implement rather than extend other classes and, and, and.
```
lukasbestle:
```suggestion
There would be so much more to learn and explore, but hopefully this recipe has lowered the entry barrier a bit, so that you are prepared to dive deeper yourself. If you explore the Kirby core classes, you will find quite a few other OOP related terms like traits, abstract methods and classes, classes that implement interfaces rather than extend other classes and, and, and.
```

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,482 @@
...
+
+OOP uses the same paradigms for data structures in a program. The properties and methods of an object like the book are described in a class, which acts as the model or blueprint for every book you might have in your application, whereas an object is an instance of this class, i.e. a specific embodiment of the class (the specific book in your library).
+
+In a larger software program, you typically have multiple classes whose objects can interact with each other (like a book, a bookshop, the customers that buy the books in your shop, invoices for sold books etc).
+
+### A simple class example
+
+Time for the first coffee. Let's start with a simple example and do some coding to see how we can actually create such a class in PHP and how we can create (instantiate) objects from that class. We define a class called `Book` with some properties that should apply to all objects of this class, and methods to work with the data of these objects.
+
+In your webserver's (glossary: document-root text: document root), create a new folder `book` and inside of this folder a new file called `Book.php` with the class definition code from above. Copy the code below into this file.
```
lukasbestle, line 51: > ⚠️ Glossary term “document-root” not found.

content/1_docs/2_cookbook/3_templating/0_understanding-oop/recipe.txt
```diff
@@ -0,0 +1,485 @@
...
+No images, lots of text 😉. But at least some code to play with. So get some coffee or tea.
+</warning>
+
+## Prerequisites
+
+If you want to code along with this recipe, you need the following:
+
+- A web server (ideally on your (link: docs/cookbook/setup/development-environment text: local machine))
+- A current version of PHP
+- A Kirby installation, either (link: docs/guide/quickstart text: Starterkit or Plainkit)
```
lukasbestle: Why is a Kirby installation required for the tutorial? As far as I can see, all classes work on their own.

texnixe: Oops, you are right, that's a leftover from the first version

lukasbestle (approved):

PR #1166, opened 2021-01-13
fix: update starterkit image

pReya:
* update the starterkit screenshot to show the latest kirby version


pReya:
@bastianallgeier Anything I can improve to get this merged?

lukasbestle:
Thank you for your PR, sorry we missed it.

Why was the image resolution changed? Apart from that, it's great!

pReya:
> Thank you for your PR, sorry we missed it.
> 
> Why was the image resolution changed? Apart from that, it's great!

I just cropped a nice looking area, without trying to match the previous width/height and then optimized the image for file size. GitHub will display it at 100% width anyway, so I thought it doesn't really matter. Do you want me to match the previous size exactly?

lukasbestle (approved):
You are right, the format and size don't really matter. I just tested it in the Starterkit README locally and I think it looks great.

pReya:
@lukasbestle When I visit https://github.com/getkirby/starterkit I still see the old image, even though the URL in the file is correct and points to the new image. Seems to be some GitHub caching magic. Can you reproduce this, too?

lukasbestle:
Yep. If you access the image URL from the README directly, it works. It can take quite long for GitHub to update their cache.

pReya:
@lukasbestle Turns out, you can actually purge the cache manually, as described here: https://docs.github.com/en/github/authenticating-to-github/about-anonymized-image-urls#removing-an-image-from-camos-cache

Just did this, and now the new image is showing up.

PR #1170, opened 2021-01-13
Rename some templates, remove unused

distantnative:
- [x] Review and approve
- [ ] Find a good time window to merge -> `develop` -> `master`
- [ ] Run script and commit actual renamed content files
- [ ] Merge

## To review, run

```
./scripts/docs-rename
```

We can add the actual renamed content files as commit once this PR has been reviewed. Leaving them out to make the PR easier to read.

## Caution ⚠️

Once merged into `develop` we should quickly merge that into `master` too.
Otherwise content changes on `master` and the remaining will diverge quickly into unsolvable merge conflicts.
So let's merge this PR only once we know we can quickly merge into `master` too.

## What it does

- Renames templates
```
    'guide'             => 'guide.chapter',
    'documentation'     => 'guide',
    'cookbook-category' => 'cookbook.category',
    'recipe'            => 'cookbook.recipe',
    'archive'           => 'versions',
    'v35'               => 'release.35',
    'issue'             => 'kosmos.issue',
    'plugin-developer'  => 'plugin.developer',
    'glossary-entry'    => 'glossary.entry'
```
- Deletes unused templates/controllers/asset files
```
- docs
- lightbox
- news
- page
- text
- upgrade
```


afbora:
`/assets/css/templates/issue.css` should be renamed `/assets/css/templates/kosmos.issue.css`

distantnative:
@afbora Great catch, added!

afbora:
@distantnative Everything looks great. It passed my tests 😉 ✅

texnixe:
Couldn't find any issues either.

afbora (approved):

distantnative:
@afbora @texnixe 
Thanks. What are your thoughts on this:
I went for rather technical names, e.g. `guide` and `guide.chapter`). Mainly to have templates that are related placed next to each other in the file list. Cause currently I find it sometimes hard to understand what templates build subunits...

But we could also try to achieve the same with simpler names, e.g.
- `guides` and `guide`
- `recipes` and `recipe`

afbora:
@distantnative I really like the idea. It's definitely clearer 👍

distantnative:
Though I just realized the limits of the simpler approach again:
If it's `recipes` and `recipe` (for `cookbook` and `recipe`) - what will `cookbook-category` become?

afbora:
How would that be? Does it make sense to you? `cookbook - recipes - recipe`  🤔 

````
- cookbook
- - recipes A
- - - recipe A1
- - - recipe AR2
- - recipes B
- - - recipe BR2
- - - recipe BR2
- - - recipe BR3
- - recipes C
- - - recipe C1
````

distantnative:
But then we have the problem as it is now again: `cookbook` and `recipe` are far apart in the template list, if you don't know it it's hard to find out that they go together...

I think our goal for ourselves should be that the site setup becomes more intuitive, easier to understand without needing some trained knowledge to connect the dots how it works.

distantnative:
Maybe we leave it as suggested in this PR - less pretty, but much clearer. And in line with the reference refactor where all reference templates get prefixed with `reference.`

afbora:
Two more ideas if you interest:

### Simple
````
recipes > recipe/s.category > recipe
````

### Advanced-Functional
````
recipes.index > recipes.category > recipes.item
````

texnixe:
@distantnative I think the names you have chosen are perfectly ok, technical or not, they make the relationships clearer.

bastianallgeier:
Why does this delete the lightbox assets? They are used in the v35 release microsite.

distantnative:
The assets? That must be a mistake then. I found an usused light box template that I deleted but the assets should stay ofc

bastianallgeier:
@distantnative sorry, I was to quick with my review. It doesn't really delete the assets. It's really just the templates. But I still found one remaining issue. The config/algolia.php also needs to be adjusted.

distantnative:
Fixed :)

bastianallgeier (approved):

bastianallgeier:
I think we should not mix this with the develop branch at all to be honest. The develop branch already has quite a few things that we have to test further before we can merge it with master and this one here seems to be pretty much independent from that. So I think it would be better to get this into master as soon as possible without the detour over the develop branch. What do you think?

distantnative:
Actually yes, that's probably for the best. Adding them both to master and develop directly

distantnative:
But since this was created from `develop` - we would first need to merge the current `develop` state into `master` and then directly this PR.

lukasbestle:
I think @bastianallgeier meant that we should rebase this PR so that it can be merged into `master` directly without having to integrate `develop` into `master` at all.

I looked into that and it would be really difficult as this PR uses the rewritten `meta` plugin from `develop` for example.
So I think the best way would be to test the current state of `develop` first so we can then safely merge everything into `master` before we continue with more PRs.

lukasbestle (commented):
I'm not sure if we should be using dots as separators. The templates would get mixed up with content representations, which would lead to two issues:

- It's not immediately clear in the template file overview which template is a representation and which is an actual template. It only becomes clear by the extension names.
- There could be unexpected issues when users access pages with content representations like when a plugin page is accessed with `.developer` in the URL or when the cookbook overview is accessed with `.recipe` in the URL. This won't happen in normal use, but it may cause vulnerabilities.

My suggestion is to use dashes instead. The grouping will still be there, but the names will be unique.

distantnative:
Would also apply to the Reference PR

distantnative:
To be honest, when I redo the names, I could do it for `master` directly. Almost same work involved. While I prefer the dots aesthetically, I totally get your points.

Will open a new PR

PR #1173, opened 2021-01-15
Refactor search: remove dependencies

distantnative:
Refactors the `search` js component by removing the `algolia` and `awesomplete` libraries


bastianallgeier (commented):
I managed a few times to create this situation where the results are somehow screwed up …

Screenshot 2021-01-21 at 14 47 23

I was also wondering if we do some kind of debouncing and caching for the search as every hit of the API will actually cost us credits. 

But except of the glitch, it works really well. I will already merge it on develop and we can do the other fixes in separate tasks.

distantnative:
@bastianallgeier Let's make search caching a separate project, since this wasn't Part of the implementation so far

And I haven't ran into this glitch myself. Any hints on how to reproduce?

bastianallgeier:
@distantnative yep, caching can be totally separated. 

It sounds a bit stupid, but it seems like typing quite fast can cause it.

PR #1174, opened 2021-01-15
Add load more with Ajax recipe and content rep glossary entry

texnixe:


bastianallgeier:
Finally found the time to read it. This is awesome! Just one question. Is there a reason you solved this with offset and limit instead of pagination?

texnixe:
I thought about using pagination but then decided to stick with the example I used before for the Kirby 2 version of this recipe. Do you think that would be the better option? I couldn't really see an immediate advantage but maybe I didn't think hard enough. I can easily rewrite it.

bastianallgeier:
I think the only argument would be simplicity. With pagination you only have to count up the page and you no longer need to calculate offset. You can also use `$pagination->hasNextPage()` for the $more var. 

It's totally fine like this. It could just be a bit more simple. I'm happy with both ways. You decide if you want to rewrite it. Otherwise just go ahead and merge. It's a great article!

Issue #1070, opened 2020-11-26
Code toolbar button preventing selection


moevbiz:
When trying to select only a fragment of a one-liner `code-toolbar` for copy-pasting, the "Copy"-button sometimes gets in the way:

<img width="615" alt="Screenshot 2020-11-26 at 10 20 35" src="https://user-images.githubusercontent.com/38008341/100333026-65a6cd00-2fd2-11eb-912c-755fe772bdef.png">
<img width="620" alt="Screenshot 2020-11-26 at 10 20 48" src="https://user-images.githubusercontent.com/38008341/100333032-66d7fa00-2fd2-11eb-89e8-836cbebcce16.png">

A `padding-right: 100px` on `.code>code` could be added to prevent this.

<img width="563" alt="Screenshot 2020-11-26 at 10 29 07" src="https://user-images.githubusercontent.com/38008341/100333612-18772b00-2fd3-11eb-8904-ee740800c383.png">
<img width="561" alt="Screenshot 2020-11-26 at 10 29 34" src="https://user-images.githubusercontent.com/38008341/100333614-190fc180-2fd3-11eb-9a0b-0b2615e8a992.png">

While at it, moving the `margin-bottom` from the `code` element to its parent `.code-toolbar` prevents the box shadow on the "Copy" button from overflowing, though I don't know if this might have unwanted layout changes somewhere else.

<img width="552" alt="Screenshot 2020-11-26 at 11 35 32" src="https://user-images.githubusercontent.com/38008341/100340444-832c6480-2fdb-11eb-8d2f-c509d6357ec5.png">

distantnative:
We are currently in the process of refactoring our JS and CSS setups. Once we have that done, we'll take a look at this 👍

distantnative:
@moevbiz The suggested padding-right fix works good for one-line code blocks. But for many lines where not the firs tone is the longest, it looks very odd that a scrollbar gets added for just scrolling into more black background.

moevbiz:
You're right. Maybe it would make sense to add this only to one-liners, since a block with many lines in most cases begins with something like `<?php` – not an overflow right in the first line.

distantnative:
We'll have a solution for this soon

PR #1180, opened 2021-01-17
Code: fix overflowing shadow on Copy #1070

distantnative:
Fixes this:
100333614-190fc180-2fd3-11eb-9a0b-0b2615e8a992

### DIFF
--- a/www/assets/css/legacy.css
+++ b/www/assets/css/legacy.css
@@ -757,7 +757,6 @@ h6 code,
   font-size: 14px;
   letter-spacing: 0;
   line-height: 1.7142857143;
-  margin-bottom: 1.5rem;
 }
 
 .code ::selection,
@@ -1017,6 +1016,7 @@ h6 code,
 :not(.codeblock-figure) > .code-toolbar {
   overflow: hidden;
   position: relative;
+  margin-bottom: 1.5rem;
 }
 
 .code-toolbar .toolbar {


afbora:
This issue seems to continue for me. Is it normal for it to look like the screenshot?

Ekran Resmi 2021-01-17 17 25 45

distantnative:
It's not ideal but better than originally:
- originally the shadow would reach outside of the code element
- in your case it is actually still contained in the code element, just that the code element itself has a scrollbar and it overlaps the scrollbar. 

Not sure how we could fix your case too.

afbora:
What if we prevented the selection of text with CSS?

````css
-webkit-touch-callout: none; /* iOS Safari */
  -webkit-user-select: none; /* Safari */
   -khtml-user-select: none; /* Konqueror HTML */
     -moz-user-select: none; /* Old versions of Firefox */
      -ms-user-select: none; /* Internet Explorer/Edge */
          user-select: none; /* Non-prefixed version, currently supported by Chrome, Edge, Opera and Firefox */
````

lukasbestle:
@afbora Do you mean on the "Copy" button? Sure, why not. :)

lukasbestle:
Regarding the shadow: Maybe we could use a less aggressive shadow that doesn't reach as far.

afbora:
> Do you mean on the "Copy" button? Sure, why not. :)

@lukasbestle Yes on the copy button 😉

distantnative:
@afbora oh yes that we can do! I thought you meant the shadow, not the selection.

@lukasbestle As far as I understand @bastianallgeier or @fabianmichael added the gradient not for the copy button but to fade out any text lines reaching outside the box, to not be clipped directly at the border but also hint a bit more towards that there's more when scrolling right.

fabianmichael:
@distantnative The shadow was in fact placed there to prevent any collisions with the copy button. I don’t know if the bug was introduced with a code/design change or if I just forgot to test with single-liners, when we started. I tested different shadow sizes for the button backdrop and this one resulted in the most subtle look – at least as long as the shadow does not exceed the code block. 

So if you got a better solution (maybe a black outline around the button would also work), please go ahead!

lukasbestle:
Maybe we could combine an outline with a smaller shadow (smaller blur radius).

distantnative:
Maybe we should solve this in the new CSS setup then that Bastian is working on instead. And leave it in the current/old code for now.

lukasbestle:
Yep, I agree. It doesn't make sense to work on this twice.

distantnative:
I think I would still merge this PR, it makes the visual bug slightly better for now.

PR #1181, opened 2021-01-17
GitHub Actions workflow for Algolia

lukasbestle:
This PR is based on `master` and not `develop` so it can be merged immediately without being blocked by `develop`.

**⚠️ Important:** Only merge this PR *after* the deployment setup was updated to stop running the script directly on the server. If this workflow and the indexing on the server would run in parallel, they would corrupt the search index on each deployment.

### DIFF
--- a/.github/workflows/algolia.yml
+++ b/.github/workflows/algolia.yml
@@ -8,6 +8,7 @@ on:
     paths:
     - content/**.txt
     - site/config/algolia.php
+    - site/models/**
     - site/plugins/**
 
 jobs:
--- a/.github/workflows/algolia.yml
+++ b/.github/workflows/algolia.yml
@@ -1,0 +1,28 @@
+# Updates the Algolia search index on all content changes
+
+name: Update Algolia index
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - content/**.txt
+    - site/config/algolia.php
+    - site/plugins/**
+
+jobs:
+  algolia:
+    name: Update Algolia index
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Cancel previous workflow runs
+        uses: styfle/cancel-workflow-action@0.7.0
+
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Update the Algolia index
+        run: scripts/algolia-index
+        env:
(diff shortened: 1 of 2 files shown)


distantnative (approved):
LGTM

.github/workflows/algolia.yml
```diff
@@ -0,0 +1,28 @@
...
+
+name: Update Algolia index
+on:
+  push:
+    branches:
+    - master
+    paths:
+    - content/**.txt
+    - site/config/algolia.php
+    - site/plugins/**
```
afbora: Just wondering if we should add a few more folders.

````
- site/config/*
- site/models/*
- site/controllers/*
- site/templates/*
````

Or whole `site` :))
````
- site/**
````


For example, algolia indexes will not be affected when the following scenarios occur 🤔 

- a new handled data may come with the model (site/models)
- a new section can be added to the template (site/templates)
- a new route or redirection can be added from the config (site/config)

lukasbestle: Templates and routes don't affect the Algolia indexing. The plugin just loops over all pages in the `$site->index()` and collects the field values, some of them run through KirbyText.

I have added the models as these could in theory indeed affect the indexing result.

afbora (approved):

.github/workflows/algolia.yml
distantnative: @lukasbestle wouldn't template changes affect it too (e.g. if we decide to add another field)?

lukasbestle: @distantnative No, the fields to use are only configured in `site/config/algolia.php`.

distantnative: @lukasbestle True, thought for a moment it'd actually parse the pages, but you're right it parses just the content files based on the config.

bastianallgeier:
Where would I need to store the secret Algolia key?

lukasbestle:
@bastianallgeier Nowhere, I've already done it. It's in the repo settings.

bastianallgeier (approved):

bastianallgeier:
I've disabled the deployment webhook and will try this now. Wish me luck 🤣

bastianallgeier:
Looks like it's not automatically running after the PR

PR #1186, opened 2021-01-20
Image compression workflow: ignore `kirby` folder

distantnative:


lukasbestle:
I don't think this will work. It will only control when the workflow will be executed, but if it executes because of changed images in the `content` dir, it will still compress images everywhere.

See https://github.com/calibreapp/image-actions#ignore-paths.

distantnative:
Added a restriction on the step as well.

distantnative:
Have you found if there's a good way to test GitHub Actions locally?

.github/workflows/images.yml
```diff
@@ -22,7 +23,7 @@ on:
       - '**.jpeg'
       - '**.png'
       - '**.webp'
-  workflow_dispatch:
```
lukasbestle, line 25: If this was removed on purpose, please also update the comment at the beginning of the file. That also still mentions that the workflow is run on schedule.

lukasbestle:
> Have you found if there's a good way to test GitHub Actions locally?

There is https://github.com/nektos/act, but I haven't tried that as I don't want to set up Docker on my Mac. It also doesn't seem to replicate the GitHub Actions feature set 100 %. So I have used test branches directly on GitHub so far.

lukasbestle (approved):

PR #1189, opened 2021-01-22
Fix cookbook I18n method reference

renestalder:
Resolve an issue in the cookbook about I18n placeholders, where the link to the reference
page is broken due to a missing forward slash


bastianallgeier (approved):

bastianallgeier:
Thanks for the fix!

PR #1194, opened 2021-01-24
Fix asset caching

distantnative:
I tried to put together what I think make sense to fix asset caching:
- Remove old Cachebuster code based on Laravel Mix (was overlooked before)
- Added `Cache-Control ` statement to `.htaccess` which should browsers make check with server for outdated JS files on each load (here more expertise would be welcome!)
- Reduced number of JS files to be loaded by
  - Inlining small JS in the templates directly (instead of template-specific JS file)
  - Integrating loader utils into the other existing JS files


bastianallgeier:
It all looks good to me except the cache header settings. I really cannot tell if they make any sense. On the first sight it looks to me as if they would make sure that none of the JS files get cached at all. @lukasbestle @afbora do you have more insight here?

afbora:
I do not know well about this but I tested the repo yesterday, it looked great 👍

distantnative:
So as I posted on Discord, the `no-cache` header actually doesn't mean it's not caching but it means that the browser needs to check with the server on each request whether there's a new file, if so the new file is send by the server if not just an empty response and the browser takes the cached file.

distantnative:
But what's missing from this PR is the config file for staging. Currently the local setup is still delivered on staging as well

bastianallgeier:
Here's the staging config file that I use on the server: 

```php
<?php

return [
    'debug' => true,
    'cache' => [
        'pages' => [
            'active' => true,
            'type'   => 'file'
        ],
        'plugins' => [
            'active' => true,
            'type'   => 'file'
        ]
    ],
];
```

distantnative:
Added the config for staging and altered the `index.php` to also do the remote mode for staging.
I think this should work, or?

distantnative:
@bastianallgeier When we merge to try this, the deploy job to staging needs to run `nom run build` - so we can test this with staging first as well before doing it on the live site.

bastianallgeier:
@distantnative looks like it should work. I will try it now.

PR #1198, opened 2021-01-30
add option 'read' to the reference of blueprints.users.permissions.pages

dennisbaum:
after a while i found out, that even more permissions are listet in the [guides-section](https://getkirby.com/docs/guide/users/permissions), than in the reference.
maybe the reference should also link to that site?


bastianallgeier (approved):

bastianallgeier:
That's great! Thanks for the fix.

PR #1184, opened 2021-01-18
Rename templates

distantnative:
Renaming templates as follows

```
    'documentation'      => 'guides',
    'cookbook-category'  => 'cookbook-category',
    'recipe'             => 'cookbook-recipe',
    'archive'            => 'versions',
    'v35'                => 'release-35',
    'issue'              => 'kosmos-issue',
    'plugin-developer'   => 'plugin-developer',
    'glossary-entry'     => 'glossary-entry',

    'method'             => 'reference-classmethod',
    'class'              => 'reference-class',
    'component'          => 'reference-component',
    'endpoint'           => 'reference-endpoint',
    'endpoints'          => 'reference-endpoints',
    'extension'          => 'reference-extension',
    'extensions'         => 'reference-extensions',
    'field-method'       => 'reference-fieldmethod',
    'field-methods'      => 'reference-fieldmethods',
    'helper'             => 'reference-helper',
    'helpers'            => 'reference-helpers',
    'hook'               => 'reference-hook',
    'hooks'              => 'reference-hooks',
    'icon'               => 'reference-icon',
    'icons'              => 'reference-icons',
    'kirbytag'           => 'reference-kirbytag',
    'kirbytags'          => 'reference-kirbytags',
    'packages'           => 'reference-packages',
    'section'            => 'reference-panelsection',
    'sections'           => 'reference-panelsections',
    'cheatsheet'         => 'reference',
    'root'               => 'reference-root',
    'roots'              => 'reference-roots',
    'cheatsheet-section' => 'reference-section',
    'cheatsheet-article' => 'reference-article',
    'ui'                 => 'reference-ui',
    'url'                => 'reference-url',
    'urls'               => 'reference-urls',
    'validator'          => 'reference-validator',
    'validators'         => 'reference-validators'
```

⚠️ What's next

- [x] review and approve (first run `./scripts/rename`)
- [x] make sure there are no new changes on `master`. If there are, rebase
- [x] run `./scripts/rename`, remove script and commit renamed content files
- [ ] merge

PR #1210, opened 2021-02-05
Rename templates

distantnative:
@bastianallgeier I think I found the issue you ran into. Haven't encountered more problems.


bastianallgeier:
What was the problem?

distantnative:
@bastianallgeier Your problem https://github.com/getkirby/getkirby.com/pull/1184#issuecomment-772458220 was probably cause but he parent page model creating the children from reflection but still assigning the old template name. At least that's a bug I found, not sure from your description what you encountered exactly.

PR #1214, opened 2021-02-10
feat: add query template to form field placeholder description

pReya:
* add query template hint to form fields parameter description (see https://github.com/getkirby/kirby/pull/2029/files)

### DIFF
--- a/kirby/src/Form/Field.php
+++ b/kirby/src/Form/Field.php
@@ -177,7 +177,7 @@ class Field extends Component
                     return I18n::translate($label, $label);
                 },
                 /**
-                 * Optional placeholder value that will be shown when the field is empty
+                 * Optional placeholder value that will be shown when the field is empty. Supports query templates like `{{ page.title }}` (since 3.2.5)
                  */
                 'placeholder' => function ($placeholder = null) {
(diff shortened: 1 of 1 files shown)


pReya:
Hmm, I just discovered, this is now supported on `before`, `after`, `default`, `label` and `placeholder`. Should I add the sentence to all these params? Or will this look repetitive? Is there a better way to indicate that query templates are supported on these params?

texnixe:
I'm sorry, this is the wrong repo for this PR, this need to go into the kirby repo.

pReya:
Sorry, I never looked at the structure of the documentation repo. Now I see that it just uses a regular Kirby instance and uses the comments from the Kirby folder. I'll submit a PR to the other Repo. But since you're the documentation queen @texnixe, do you have any advice on my question? Is it a good idea to copy & repeat the given sentence after every param that supports query templates? Or is there a better way?

distantnative:
I think we're starting to repeat this too often then. Probably rather need to add a general part somewhere about this feature. What do you think @texnixe?

distantnative:
Closing this for the other approach we discussed on Discord

PR #1227, opened 2021-02-21
Custom block type recipe

texnixe:
Hey guys, could you for a start have a look if the final code is ok, what is missing or could be done more easily?


bastianallgeier:
AWESOME! I just tried to create the plugin myself but ran into an issue with the block blueprint and I cannot figure out where the issue is. The field definition is simply not being loaded. No matter what I try. Do you have the final version of the plugin lying around and if yes, could you send it to me?

bastianallgeier:
After realising that I'm super stupid and added all the example files for the plugin directly in the plugins folder instead of a audio-plugin subfolder, I'm back on track and will follow the rest of the tutorial on Monday. But so far it's all perfect and I only realized some smaller things. 

I.e. after creating the very first basic steps of the plugin, I think there should be a little excerpt how to use the new audio block in the blocks field afterwards. 

More feedback will follow.

distantnative:
We changed the template names yesterday, so the new one is `cookbook-recipe`.

bastianallgeier:
Ok, I'm through with the whole tutorial and there are a few things I would change. 

I just wonder how to add them in the best way. I've modified the final result in a way that would fit better to our current style of creating Panel components. I don't want to mess around with the article too much. It's mostly about the part how to fetch file object data and a few styling things. If you want, I can take over and add them. 

Here's my final component: 

```html
<template>
  <k-block-figure
    :is-empty="!sourceUrl"
    empty-icon="audio-file"
    empty-text="No file selected yet …"
    @open="open"
    @update="update"
  >
    <div class="k-block-type-audio-wrapper">
      <div>
        <k-aspect-ratio
          class="k-block-type-audio-poster"
          cover="true"
          ratio="1/1"
        >
          <img
            v-if="posterUrl"
            :src="posterUrl"
            alt=""
          >
        </k-aspect-ratio>
      </div>
      <div @dblclick.stop>
        <k-writer
          :inline="true"
          :marks="false"
          :placeholder="field('title').placeholder"
          :value="content.title"
          class="k-block-type-audio-title"
          @input="update({ title: $event })"
        />
        <k-writer
          :inline="true"
          :marks="false"
          :placeholder="field('subtitle').placeholder"
          :value="content.subtitle"
          class="k-block-type-audio-subtitle"
          @input="update({ subtitle: $event })"
        />
        <k-writer
          :inline="true"
          :marks="false"
          :placeholder="field('description').placeholder"
          :value="content.description"
          class="k-block-type-audio-description"
          @input="update({ description: $event })"
        />
        <audio class="k-block-type-audio-element" controls>
          <source :src="sourceUrl" :type="mime">
        </audio>
      </div>
    </div>
  </k-block-figure>
</template>

<script>
export default {
  data() {
    return {
      mime: ""
    };
  },
  computed: {
    poster() {
      return this.content.poster[0] || {};
    },
    posterUrl() {
      return this.poster.url;
    },
    source() {
      return this.content.source[0] || {};
    },
    sourceUrl() {
      return this.source.url;
    },
  },
  watch: {
    "source.link": {
      handler (link) {
        if (!link) return;

        this.$api.get(link).then(file => {
          this.mime = file.mime;
        });
      },
      immediate: true
    }
  }
};
</script>

<style lang="scss">
.k-block-type-audio-wrapper {
  display: flex;
  padding: 1rem;
  background: black;
  color: white;
}
.k-block-type-audio-poster {
  width: 12rem;
  margin-right: 1rem;
  background: #333;
}
.k-block-type-audio-title,
.k-block-type-audio-subtitle {
  font-size: 1.5rem;
}
.k-block-type-audio-title {
  font-weight: 600;
}
.k-block-type-audio-subtitle {
  margin-bottom: 1rem;
  color: #999;
}
.k-block-type-audio-description {
  line-height: 1.5;
}
.k-block-type-audio-element {
  margin-top: 2rem;
  height: 2rem;
}
</style>
```

texnixe:
@bastianallgeier Thanks for checking it out. Yes, please go ahead.

lukasbestle:
Is it intended that this PR changes 1050 files? Or is this an unintended consequence of the template name change and needs to be fixed by rebasing this PR against `master` again?

distantnative:
@lukasbestle I think a rebase is needed

bastianallgeier:
I just rebased it.

PR #1236, opened 2021-03-05
Suggest values for `step.unit` option in time field

moritzebeling:
It is not really obvious how to set the time field to seconds and you would have to guess, that you can do
```yaml
step: second
```
To make that more clear, I suggest to add:
> Possible values for unit are `hour`, `minute` and `second`.


texnixe:
Thank you!

PR #1239, opened 2021-03-06
feat: add info for unremovable draft status in page blueprints

pReya:
* add info that the `draft` status of a page can not be removed or disabled. so far the description suggests that all statuses can be removed – which is not the case and can be very misleading


texnixe:
Thanks, Moritz, for adding this information!

PR #1261, opened 2021-04-10
Reference: automatic default files for blocks

### LINKED ISSUE #1230: Blocks: outdated snippets and blueprints
Each block's source (snippet and blueprint) is pasted manually into the content file. For example: [1_docs/3_reference/3_panel/5_blocks/4_image/reference-article.txt](https://github.com/getkirby/getkirby.com/blob/master/content/1_docs/3_reference/3_panel/5_blocks/4_image/reference-article.txt)

A lot of the default source code is outdated already, e.g. the image block has been updated lately.

Wouldn't it be simpler to the include the source directly from [kirby/config/blocks](https://github.com/getkirby/kirby/tree/master/config/blocks), similar to parsing PHPDoc? Thus the docs will be future-proof and always represent latest fixes and changes.

distantnative:
Fixes https://github.com/getkirby/getkirby.com/issues/1230


site/templates/reference-block.php
```diff
@@ -0,0 +1,31 @@
...
+  <h2 id="preview"><a href="#preview">Preview</a></h2>
+  <?= kirbytag('screenshot', $page->image()->filename()) ?>
+
+  <?= $page->text()->kt() ?>
+
+  <h2 id="default-files"><a href="#default-files">Default files</a></h2>
+
+  <h3 id="default-files__snippet"><a href="#default-files__snippet">Snippet</a></h3>
+  <?= kt('
+```php "/site/snippets/blocks/' . $page->slug() . '.php"
```
texnixe: I'd remove the file location, because the default file doesn't exist there (there is also no location given for the blueprint).

texnixe (approved):
Apart from the one comment, it's great.

lukasbestle (commented):
I agree with Sonja.

distantnative:
I think Bastian added the file path in the original docs to highlight where the user needs to put a customised snippet to overwrite the default. So if we don't include it as line above the code, we should still add a comment below the code block then.

lukasbestle:
That makes sense. I'd be fine with both variants (file location in the code block or an info box at the end), but in any case it should be consistent between the snippet and blueprint.

distantnative:
Added it in the code block for the blueprint as well then :)

texnixe:
Wouldn't surprise me if people went looking in their local folders to find the files there and then be irritated that it is not there. I'd prefer a comment where to overwrite the files.

distantnative:
@texnixe better?

texnixe:
@distantnative Perfect!

texnixe (approved):

PR #1268, opened 2021-04-15
Only highlight Kirby classes with reference link

distantnative:
Have been wondering whether it is better to only add the yellow color to classes from Kirby that then also link to the Reference. So leave all generic classes gray. 

Old:
Screen Shot 2021-04-15 at 09 22 08

New:
Screen Shot 2021-04-15 at 09 22 17

Would simplify the parsing. But unsure myself.

### DIFF
--- a/site/plugins/site/src/Types/Type.php
+++ b/site/plugins/site/src/Types/Type.php
@@ -105,27 +105,17 @@ class Type
             return static::tag($type, $type);
         }
 
-        // Assume, it’s a class name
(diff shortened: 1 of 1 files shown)


lukasbestle:
I think that makes sense. Otherwise you expect that you can click on it like with the Kirby classes.

distantnative:
Actually, this isn't fixing the other issue I had in mind.

And maybe it'd be also odd to have them grey like `null` while other normal and unclickable types (string, int...) still have colors?

PR #1271, opened 2021-04-15
Loading indicator for search

distantnative:
Is this what you thought of?

Would need some cleaning up - but just to make sure that's what you meant.


bastianallgeier:
Perfect! I just added a timeout to not show the indicator immediately.

bastianallgeier:
Debouncing already fixed the slowness for me.

PR #1304, opened 2021-04-16
Refactor banner implementation

### LINKED ISSUE #1290: `banner` config: remove `enabled`
@lukasbestle @afbora couldn't we remove the `enabled` key and just base it on the start and end dates?

### LINKED ISSUE #1291: Sales banner throughout the site
Shouldn't we show it throughout the side (maybe not Reference) and not just on home?

With the automatic dates, this could be a caching problem.
But isn't that already on home?

### LINKED ISSUE #1296: Sale banner: Automatic cache flushing
I think we can solve this by giving the homepage an automatic cache expiry date based on the configured dates.

lukasbestle:
## Status: ⚠️ Requires Kirby 3.5.5

Closes #1290.
Closes #1296.

Uses the Kirby PR https://github.com/getkirby/kirby/pull/3246.


site/plugins/site/helpers/banner.php
```diff
@@ -51,7 +52,11 @@ function banner(): ?Obj {
...
     if ($banner !== null && $banner->endDate()) {
-        $expires = min($expires, $banner->endDate());
+        if (is_int($expires) === true) {
+            $expires = min($expires, $banner->endDate());
+        } else {
+            $expires = $banner->endDate();
+        }
     }
 
     kirby()->response()->cacheExpiry($expires);
```
distantnative: Do we need to check here if `$expires` has a value?

afbora: I guess no need. `$expires` default value is `null`, and `cacheExpiry()` accepts `null`.

lukasbestle: If we pass `null`, the method behaves like a getter, so it will work as well.

distantnative: Ok 👍

distantnative (approved):
LGTM.

I would wait though for @bastianallgeier's opinion on the mobile styling.

afbora (approved):
Everything looks great for me too 👍

bastianallgeier:
As much as I like this solution, I really don't want a dev version running our site. Not because of stability issues, but because of the message it sends. We should be able to get our stuff done like any other user without hacking the core and otherwise wait for the release.

distantnative:
@bastianallgeier haven't we used RC etc. before as well? I don't think this shouldn't be an issue. We are not hacking it, just using a feature from the next release.

bastianallgeier:
No, I don't remember that we ever used an RC

bastianallgeier:
It's also super risky. The reference could include something that's not even available yet.

distantnative:
> It's also super risky. The reference could include something that's not even available yet.

That's a very good point indeed

bastianallgeier:
Can we somehow extract the rest and only fix the cache problem once the release is out?

lukasbestle:
I will take a look.

distantnative:
@bastianallgeier extracted the mobile part here: https://github.com/getkirby/getkirby.com/pull/1330

site/snippets/layouts/banner.php
```diff
@@ -1,9 +1,5 @@
-<?php if (empty($banner) === false && empty($banner['text']) === false): ?>
+<?php if (empty($banner) === false && $banner->text()): ?>
 <aside class="banner text-sm">
-  <?php if (isset($banner['url'])): ?>
-  <a href="<?= $banner['url'] ?>"><?= $banner['text'] ?></a>
-  <?php else: ?>
-  <span><?= $banner['text'] ?></span>
-  <?php endif ?>
+  <a href="<?= url('buy') ?>"><?= $banner->text() ?></a>
```
distantnative, line 3: I was thinking whether shouldn't leave it more flexible, e.g. to also be able to have a banner for an upcoming live stream. But I guess you went this way due to the current styling (pointing to Buy)?

distantnative:
One thought,  but actually only thinking out loud and already answering my own question why you did it that way. Looks good to me!

site/snippets/layouts/banner.php
lukasbestle: Yep. I thought about adding a new option `isSale` to control the special placement on the buy page, bit given that the banner always points to buy, I think it doesn't make much sense to use it for other purposes besides sales.

distantnative (approved):

site/snippets/layouts/banner.php
lukasbestle: @bastianallgeier What do you think about this? It could make sense to use the banner for different purposes than sales as Nico mentioned (live streams etc.). In this case we'd need a styling variant where the banner doesn't appear next to the Buy link though. This styling variant could be controlled automatically in the snippet depending on the target URL.

bastianallgeier: I like that a lot. The banner could automatically move to a different menu item that matches or be global like on the old site. I would handle that in a different PR when we need it though.

bastianallgeier (approved):

PR #1307, opened 2021-04-17
Improve Reference responsiveness

### LINKED ISSUE #1306: Reference: main column gets too narrow on desktop layout
The main column of the three columns layout in "reference" becomes very narrow between some breakpoints. Readability and especially the “properties” table suffers from this. The table must then be scrolled horizontically.

Also, inside the table again only the right colum adapts. As this column holds text blocks, lines become very short.

It occurs between quite common screen widths of 960px and 1280px. See screenshots below.

One possible solution might be to also adjust width/layout of the middle column, not only of the main column; another, to make the table itself respond further.

Bildschirmfoto 2021-04-17 um 11 09 05

Bildschirmfoto 2021-04-17 um 11 09 25

distantnative:
Fixes #1306

### DIFF
--- a/assets/css/layouts/reference.css
+++ b/assets/css/layouts/reference.css
@@ -24,6 +24,8 @@ html {
   width: 2rem;
 }
 .reference-panels {
+  --sidebar: 12rem;
+  --entries: 18rem;
   display: flex;
   flex-grow: 1;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


bastianallgeier (approved):

bastianallgeier:
✅

distantnative:
@bastianallgeier you meant to merge this instead close?

bastianallgeier:
WTF! Yes, of course.

PR #1309, opened 2021-04-17
Show ongoing sale on `buy` page

### LINKED ISSUE #1292: Show ongoing sale on `buy` page
To make also clear that the currently displayed price isn't the regular price. Avoiding bad surprises if someone saw the price the first time during a sale, didn't realise it and comes back after the sale.

distantnative:
Fixes #1292

Screen Shot 2021-04-17 at 15 38 01


site/templates/buy.php
```diff
@@ -86,30 +93,37 @@
...
+    const listPrice    = product.list_price.net;
+    const isSale       = currentPrice !== listPrice;
 
-  Paddle.Product.Prices(499826, function(prices) {
-    let price = prices.price.net.replace(".00", "").replace("US$", "$");
-    let priceElement = document.querySelector(".price");
-    let vatElement = document.querySelector(".vat");
+    const formatter = (price) => {
+      return new Intl.NumberFormat("en", { style: "currency", currencyDisplay: "narrowSymbol", currency: currency, minimumFractionDigits: 0 }).format(price)
+    };
```
lukasbestle: I think `formatter` should just be set to the `NumberFormat` object and then the `.format()` function can be called on it below. Spawning a new formatter for each price is not needed.

Also: Careful about the `narrowSymbol` display. It's not supported in every browser, e.g. current versions of Safari.

lukasbestle:
```js
let formatter;
try {
  formatter = new Intl.NumberFormat("en", {
    style: "currency",
    currency,
    currencyDisplay: "narrowSymbol"
  });
} catch (e) {
  if (e.constructor !== RangeError) {
    throw e;
  }

  formatter = new Intl.NumberFormat("en", {
    style: "currency",
    currency
  });
}
```

distantnative: That's unfortunate. That will give us long strings for many countries still then:

Screen Shot 2021-04-17 at 16 20 50

Screen Shot 2021-04-17 at 16 21 11

distantnative: GitHub only showed your first comment

PR #1313, opened 2021-04-17
`author` kirbytag

### LINKED ISSUE #884: Cookbook: Highlight contributor/author section with a special styling
Recently, I made a Medium article importer for Kirby and asked @texnixe if I could write a Cookbook recipe about it because it was a very simple showcase of a section plugin. She agreed and we also talked about Xdebug and me writing a recipe about that as well...

I want to make these recipes, but I also have a ton of work at [Oblik Studio](https://oblik.studio/). We're striving to increase our presence on the web and I'd get much more encouragement from my team to take the time for these recipes if we had our name somewhere in them. Do you think that could happen?

Reasons I think that would be great:

- Other people would be more interested in contributing to recipes as well
- If something's not clear, you could ask the author for more assistance in the forum, for example. In turn, the author could edit the recipe to make it more clear for others
- It would make Kirby more transparent - you'd see the actual people behind that content. I believe that would be beneficial for both the authors _and_ Kirby

distantnative:
Fixes #884

I don't think this styling is good yet:
Screen Shot 2021-04-17 at 21 17 43

But maybe this way we can star the conversation where exactly to include these (e.g. maybe instead directly below the title), what info should be shown etc.

And whether we think a tag to be used anywhere in the site makes sense. Or if it's a field for cookbook recipes.

### DIFF
--- a/content/docs/2_cookbook/3_templating/0_shared-controllers/cookbook-recipe.txt
+++ b/content/docs/2_cookbook/3_templating/0_shared-controllers/cookbook-recipe.txt
@@ -16,6 +16,8 @@ Controllers are an excellent tool to keep your templates and blueprints clean an
 
 Luckily for us, creating a shared controller is an easy task in Kirby 3.
 
+(author: Manu Moreale link: https://manuelmoreale.com)
+
 ## A simple example
 
 We'll use the `<title>` and `<meta name="description">` tags as an example. The easiest way to add those two tags to our site would be to add them directly into the `header.php` snippet. Something like this:
@@ -199,7 +201,3 @@ return function ($page, $pages, $site, $kirby) {
 
 };
 ```
-
-## Author
-
-This cookbook recipe was written by (link: https://manuelmoreale.com text: Manu Moreale).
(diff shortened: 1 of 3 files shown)


lukasbestle (commented):
I like the positioning after the intro paragraphs. So I think a KirbyTag would definitely be the most flexible solution, even if it's only going to be used in the cookbook.

The content (name and link) is good in my opinion. Maybe we could add the option for an author image, but on the other hand we'd need to update these images all the time and in every recipe unless we make a global store of author data. So I think the implementation is pretty good for the time being.

bastianallgeier:
I think it would be great to take this chance and do the following things: 

1. we should start adding an authors directory where we store bio, profile pic and links
2. cookbook recipes get references to the authors 
3. cookbook recipes should get proper intros. It's a shame that we don't use them there. 
4. if an author exists, we could automatically show it below the intro without having to work with a kirbytag
5. we should add us as authors to existing recipes wherever it makes sense
6. finally we should show little avatars of authors on the cookbook overview page as well to give it a human touch.

bastianallgeier:
I'm closing this in favour of: https://github.com/getkirby/getkirby.com/pull/1333

PR #1317, opened 2021-04-18
Automatic inline methods links

### LINKED ISSUE #1262: Automatically link to Reference for class methods

distantnative:
Closes #1262.

- Site now tries to parse all inline codes for linking to Reference pages for class methods (and field methods, helpers)
- We can also manually do so via tags: 
```
(class: Kirby\Cms\Language method: id)
(method: Kirby\Cms\App::template text: ->template())
```

First content changes added, more can follow later but shouldn't be a blocking factor.


afbora:
There is no problem with all changes within the `/content` directory 👍

bastianallgeier (approved):

PR #1318, opened 2021-04-18
Backups recipe

distantnative:
I have transferred @sylvainjule's initial draft into our new site.

We still need to read through the actual recipe and polish it before publishing.


texnixe:
I'ts probably best to compare it to the changes I originally suggested back in the days when this recipe was created. Hope I'll get to it sometime this week.

texnixe:
What are we going to do with this? It's a great recipe but since it's been lying around for ages, it is now outdated and would need an update to Panel areas at least, maybe more.

lukasbestle:
I feel like this recipe could become the "advanced continuation" for the "My first Panel area" recipe. I like that it really goes into detail.

The question is if it even makes sense to update it to areas. Especially since it's mostly code blocks and videos. The videos could probably be kept as they are even with an area rewrite, the code blocks are much more complex. We also don't update all older recipes every time there's a new Kirby feature and I think this recipe can still have a lot of value for the inspiration it brings.

On the other hand: If we release it now, the expectation would probably be that it shows the recommended way of doing things, which is no longer the case because of areas.

distantnative:
Closing this as there has been no activity for a year

PR #1319, opened 2021-04-18
Add XSS escaping guide

lukasbestle:


content/docs/1_guide/4_content/4_fields/guide.txt
```diff
@@ -50,7 +50,7 @@ In general, you are free to name your fields like you want to, but there are two
 Field content is accessible in your templates via the field's name:
 
 ```php
-<?= $page->title()->html() ?>
+<?= $page->title()->escape() ?>
 <?= $page->text()->kirbytext() ?>
```
texnixe: What about `kirbytext()` and `toDate()` here?

content/docs/1_guide/5_templates/8_escaping/guide.txt
```diff
@@ -0,0 +1,154 @@
...
+
+Intro: Some content you output in your templates or snippets may be untrusted. For example it may come from user-generated content like a registration or comment form or you might use content from external sources like an API or external database. When you cannot fully trust the content, it is important to escape it correctly before it's sent to the browser to prevent cross-site scripting (XSS) attacks.
+
+----
+
+Text:
+
+## What is an XSS attack?
+
+(link: https://en.wikipedia.org/wiki/Cross-site_scripting text: Cross-site scripting) (XSS) is a type of security vulnerability that allows attackers to run arbitrary JavaScript in the browser of other, legitimate users of your site. This attack can be used to extract critical authentication information like session cookies from your users' browsers or to manipulate your site in unexpected ways.
```
texnixe: browsers of other, legitimate

distantnative:
This is massive! 👏

Should we also extend it to change our snippets/templates/controllers? Since the site is public, would be a good leading example, wouldn't it? (even if most places are rather safe from malicious input)

texnixe (approved):

texnixe:
Would make sense. Although probably not as first priority, after all, we don't have user generated content that is not reviewed before merged.

distantnative:
Yea, it'd be more an "if someone copies our code to use it for their site". E.g. I think one of the only places could be search right now, but we strip tags at least there already: https://github.com/getkirby/getkirby.com/blob/main/site/controllers/search.php#L26

content/docs/1_guide/5_templates/8_escaping/guide.txt
lukasbestle: Ha! See, I can't spot my own typos either. ;)

content/docs/1_guide/4_content/4_fields/guide.txt
lukasbestle: I fixed the `kirbytext()` example. `toDate()` is fairly safe with a fixed format string.

lukasbestle:
@distantnative I agree, but it shouldn't be a blocker for this PR. I've created an issue here: https://github.com/getkirby/getkirby.com/issues/1321

bastianallgeier (approved):

PR #1335, opened 2021-05-01
UI Kit docs

distantnative:
## Status: in progress 👷

- UI Kit docs now also get (mostly) automatically generated by parsing JSDoc comments in the core

### TODO
- [x] Adapt `models/reference-uis.php` to load file from cache or remote
- [x] Set up `files.getkirby.com`
- [x] Integrate `npm run build:docs` and uploading `ui.json` into release routine of `getkirby/kirby`

### Related
- https://github.com/vue-styleguidist/vue-styleguidist/issues/1166


distantnative:
Encountering some difficulties https://github.com/vue-styleguidist/vue-styleguidist/issues/1092

distantnative:
@lukasbestle So I think the latest state of our discussion was: the JSON file lives on the website server, we have a command to easily download and add it to our local setups. Maybe you could help me think through:

1. Github action that creates `ui.json` on release of a new Kirby version and uploads it to `/assets` on the getkirby.com server
2. composer/npm script that downloads file from website to local setup

bastianallgeier:
I think I would still solve this directly in PHP.

1. add a new cache configuration for the UI kit
2. download the ui.json in PHP from the repo when the cache does not exist
3. keep it in the cache with the Kirby version as cache file name. 
4. clean the cache if the version number has changed

This way, we don't need to have any external setup for this. We can solve it all in the site repo.

distantnative:
You mean for the 2nd step? Cause the JSON file is not part of any repo, so we would need something external at least for uploading it to the website server.

And why caching it? We generate virtual pages from it that are cached anyways. Since we also have no version switcher on the website anyways, I think it makes sense to always just overwrite thee file with the most recent version.

bastianallgeier:
Oh, I thought it's part of the kirby repo but ignored in the zip. I was thinking about downloading it directly from Github. That's why the cache makes sense. Otherwise we will download it quite often after every content change.

distantnative:
We said we didn't want it to be included in any repo, so it's on `.gitignore` in both repos.
And we would only need to download it after each release.

distantnative:
We could include it in the `kirby` repo and ignore it for the ZIP. But I think I'd still rather go for a script that only gets triggered (automatically/manually) on each release than putting PHP code in the website that keeps checking Github even when no release has happened.

bastianallgeier:
If such a script is not complicated, I'm fine with it.

distantnative:
Actually just thinking: wouldn't a GitHub action that updates the getkirby.com repo with the new Kirby version on every release good anyways?

bastianallgeier:
Yes, absolutely. I think @afbora and @lukasbestle were already looking into something like that for the starterkit and plainkit as well.

lukasbestle:
A few thoughts:

- I agree with Nico that the JSON file should not be in the repos. It's quite large and because it will change with every release, the Git repo would need to include quite many diffs and would grow in size.
- I like Bastian's idea of downloading and caching the file by Kirby version. This gives us a unique cache identifier that makes sure that the correct file is present. This setup could also seamlessly be used for a multi-version reference later. But even at the moment it's still useful for more reliable caching.
- And automatic downloading in PHP is much simpler than a custom script. We need to keep in mind that the getkirby.com site is used locally by many of our users, not just by us.
- Until we have the multi-version reference, I think the file to download should just be based on the `kirby()->version()` of the getkirby.com installation.
- Do we need to upload the files directly to the getkirby.com site? Maybe we could set up a new subdomain for files like this, for example files.getkirby.com. This would then be clearly separated from the site itself, which makes GitHub Actions less risky.
- I had made some plans to automate the release process more using Actions, but to be honest it starts to feel like [xkcd Automation](https://xkcd.com/1319/). Manually uploading the JSON to the server on release isn't too bad and we can still automate this later. I don't see it as a blocker at the moment.

distantnative:
I still don't quite understand the reason for caching? Is a cache file faster than the JSON file itself? Especially considering that we generate virtual pages from it that themselves get cached anyways. Wouldn't it be much simpler to just save the file as `360.json` or so? The file itself is already a cache of the doc parser in a way.

The rest I can agree. We just can't forget to always generate it with a new release. Upload it to the server and then we just add a little PHP part to our site that downloads the file from the server if it isn't present locally.

lukasbestle:
> Wouldn't it be much simpler to just save the file as 360.json or so?

That's exactly what we want to do with caching. The file is downloaded from getkirby.com and then stored in cache.

distantnative:
Okay maybe it's semantics but if I read your reply, I understand that we download the file but then we also put it in a Kirby cache. Thereby creating two copies of that file. Why not just leave the downloaded file and use that?

bastianallgeier:
I think this is exactly what we mean. We get the JSON and save it in the cache. Done :) The cache is just a convenience because it already handles expiry. Otherwise we would need to solve that again.

distantnative:
But why do we need to solve expiration? This file should never expire, that's just unnecessary requests to load again the exact same file. For one Kirby version there'll always be only one version of the JSON file. Using the cache feels overly engineered when we could just download the file and that's it.

lukasbestle:
> I understand that we download the file but then we also put it in a Kirby cache. Thereby creating two copies of that file.

🤔 Why two copies? The file is requested, the JSON data is parsed and then the array is stored in cache. Where would the second copy be?

I agree that expiry is not needed here, but I think the main advantage of using the cache is that the cache directory is already transient.

distantnative:
We should get this released together with 3.6.0

@bastianallgeier what do you think about Lukas' idea to set up files.getkirby.com for that use case?

distantnative:
@bastianallgeier @lukasbestle Since 3.6 is near, would be great if we could this PR ready.

lukasbestle:
To be honest I feel like we don't have resources to finish this until the release. We can still do it right after the release as the next "bigger" project, or is there anything that we can only do in 3.6?

distantnative:
You think this is a big thing? I thought we just have to agree where to put the file and then write a few lines how to download it. That should be all, right?

I would consider this 3.6 docs - just like all the rest in https://github.com/getkirby/getkirby.com/pull/1420 - that really should happen with the release. Otherwise the current UI docs would be even more dramatically outdated at the 3.6 release (including components that don't exist anymore etc.).

lukasbestle:
You are right. :)

Regarding your todos in the first post: I think we still misunderstood each other. What Bastian and I meant is that the `ReferenceUisPage` model directly looks for the data in Kirby's cache and automatically downloads and stores the data in cache if it's not already there. It should not be necessary to write an additional script that does this.

Basically something like this:

```php
$cache  = $this->kirby()->cache('reference');
$uiFile = 'ui/' . $this->kirby()->version();
$data   = $cache->get($uiFile);

if (!$data) {
    $data = Remote::get('https://files.getkirby.com/' . $uiFile . '.json')->json();
    $data = array_filter($data, function ($ui) {
        // only include components that
        // have been flagged as public
        return (
            isset($ui['tags']['internal']) === false ||
            $ui['tags']['internal'] === null
        );
    });

    $cache->set($uiFile, $data);
}

foreach ($data as $ui) {
    ...
```

In parallel @bastianallgeier should set up the `files.getkirby.com` subdomain where we will store the JSON files.

distantnative:
@lukasbestle Great! Added it as suggested - currently using a local URL - once @bastianallgeier has set up the file server, we can change that URL

bastianallgeier:
Couldn't we grab it directly from GitHub?

distantnative:
Feel like we're turning in circles 😆😂 we said we didn't want to include it in the repository - if we see that different now, of course we could

lukasbestle:
I still think we should keep the file out of the `kirby` repo because it will blow up the repo size over time.
In theory we could set up a separate `getkirby/files` repo and use that for hosting. The advantage is that we can push to it directly without access to our own server, but I feel like it's a hack and not an intended use case for a repo.

bastianallgeier:
Sorry, you are right. I forgot what the last state of our discussion was. How could we solve the upload in a good way? FTP? POST request? rsync?

bastianallgeier:
OK, one more thought. With npm run build:docs couldn't we even use a shiny free Netlify account with a custom subdomain to build and host the file on demand for us? That way we wouldn't even need to think about uploading somewhere.

bastianallgeier:
I have to say that Netlify really isn't too bad for this kind of stuff. Took literally 3 minutes to get this up and running whenever a new commit is pushed to development: https://agitated-engelbart-fd3a76.netlify.app/ui.json

bastianallgeier:
Ha, even better. DeployHQ can do the exact same thing and move it to our shared hosting server without any issues. With the wildcard certificate over there, we have a really solid setup out of the box. 

https://ui.getkirby.com/index.json

So far it is redeploying with every commit on develop. We can easily change that later to main. 

I really love it. It's super simple and solid without any witchcraft in the background.

bastianallgeier:
So far it is redeploying with every commit on develop. We can easily change that later to main.

bastianallgeier:
Ok, holy shit. I just browsed the generated UI docs for the first time – I know I should have done that earlier – and it's so good! There are of course issues here and there, but in general it's just lovely!

distantnative:
If we always only use one JSON file (without Kirby version in file name), we could remove that from the cache key too.

And yes I think this is setting a good basis. There are issues with the library, issues with our code here and a lot of lacking inline docs. But it's a big step in the right direction to finally maintain docs for the UI components, I believe :)

distantnative:
Oh but also, @bastianallgeier, DeployHQ should only run with changes on `main`. Otherwise we'll publish UI component docs for upcoming versions before they are released.

distantnative:
Another thought: should we add some lines that allow providing a local JSON file instead? That way when working on online docs, we quickly could throw in the JSON file to see the results in our local getkirby.com instance

lukasbestle:
Regarding multiple versions: That would be relevant for the multi-version reference we want to build at some point, but for now a single file is enough. However the question is: Might it make sense to already "collect" version-specific files? Or would be alright to only start with version-specific UI kit docs once we have actually built that feature later (i.e. it wouldn't be possible to go back in time at that point, only forward in time).

> Another thought: should we add some lines that allow providing a local JSON file instead?

The model could automatically use the file inside the `kirby` directory if it has been built there. Then we can check out any `kirby` commit in the getkirby.com repo, build the JSON file and it will just work™️. This behavior would be in line with the rest of the reference, which also uses the current state of the local `kirby` folder.

bastianallgeier:
@distantnative I set it up to deploy on develop changes only during our RC phase. I will switch that later. 

I don't think we need to worry about multiple versions yet. 

To link to a local file is a great idea.

distantnative:
Ok added support for the local file. I think we could merge this into the main 3.6 docs branch/PR now.

PR #1339, opened 2021-05-02
Add Docker recipe

texnixe:
I have created a monster...

If anyone with a Windows device has a chance to test the last example, that would be huge.


bastianallgeier (approved):
Wow, this is incredible! I've never really dared diving deeper into Docker containers and this intro is so useful. <3

content/docs/2_cookbook/9_setup/0_kirby-meets-docker/cookbook-recipe.txt
```diff
@@ -0,0 +1,949 @@
...
+      - ./starterkit:/var/www/html/
+    env_file:
+      - ./id.env
+```
+
+### `docker-compose.yml` explained
+
+A `docker-compose.yml` file always starts with the Docker Compose version number. The current minor version at the time of writing this recipe is 3.8.
+
+After that, the next top level keyword is `services`, which is followed by an indented list of named services. The names of these services are arbitrary, but of course it makes sense to give them names that indicate their purpose. In the file above, we currently have only one service called `web server`.
```
afbora:
```suggestion
After that, the next top level keyword is `services`, which is followed by an indented list of named services. The names of these services are arbitrary, but of course it makes sense to give them names that indicate their purpose. In the file above, we currently have only one service called `webserver`.
```

afbora (approved):

bastianallgeier:
Can we merge it? @texnixe should I already promote it today through Discord and Twitter or wait until the newsletter's out tomorrow?

texnixe:
@bastianallgeier: Let's see if @lukasbestle finds the time to proof-read it this morning, if not, then we just merge it and fix potential typos later.

PR #1344, opened 2021-05-06
Bring back plugin representation

afbora:
## ⚠️ Before Merge

I added token to the `header` to avoid limit problems. ~You've to add `GITHUB_ACCESS_TOKEN` env variable.~

### DIFF
--- a/site/config/config.php
+++ b/site/config/config.php
@@ -4,7 +4,7 @@ return [
     'api'      => false,
     'banner'   => require __DIR__ . '/banner.php',
     'debug'    => true,
-    'github'   => 'https://github.com/getkirby',
+    'github'   => require __DIR__ . '/github.php',
     'features' => require __DIR__ . '/features.php',
     'meta'     => require __DIR__ . '/meta.php',
     'panel'    => false,
--- a/site/config/github.php
+++ b/site/config/github.php
@@ -1,0 +1,6 @@
+<?php
+
+return [
+    'url' => 'https://github.com/getkirby',
+    'key' => trim(F::read(__DIR__ . '/keys/github.txt'))
+];
--- a/site/models/plugin.php
+++ b/site/models/plugin.php
@@ -3,7 +3,6 @@
 use Kirby\Data\Data;
 use Kirby\Cms\File;
 use Kirby\Cms\Nest;
-use Kirby\Cache\FileCache;
 use Kirby\Toolkit\Str;
 
 class PluginPage extends Page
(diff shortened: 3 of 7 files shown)


afbora:
Also, what if we increase the version cache duration of the plugins from 1 hour (60 mins) to 3 hours (180 mins) for performance?

lukasbestle (commented):
Isn't the GitHub Access Token the user-specific one? Because we don't have a dummy/bot user at the moment, I think it isn't really an option for us to use one.

So far we got along quite well without one given our caching. :+1: on increasing the cache duration to three hours.

afbora:
I didn't know GitHub Access Token was not available in organizations. But can't we use one of us's personal token?

I'm only worried about two things:

1. If we think of it as an plugin management system in the kirby core (version control of kirby and installed plugins), the limits will become a problem.

2. I do not know the number of visitors to the website, as there will be an excess of the limit in instant traffic increases, the API will not return data and the version will be cached as `false`.

If we can use a personal token, I think it will not hurt.

afbora:
Also, the cache needs to be reset in the plugins presentation. Because all versions are `false`: https://getkirby.com/plugins.json

lukasbestle:
As long as we use caching, won't the number of API requests be limited by the number of plugins? The worst case would be that all plugins are queried at once, but that's still not an insane number of requests.

Have you found a source in the GitHub docs how high the rate limit is for unauthenticated API requests?

> But can't we use one of us's personal token?

Using one of our own tokens is not great. If that person were to leave the organization, stuff would break immediately.
If we really need a token, the best way would be to create a bot user and generate a token for that.

afbora:
> As long as we use caching, won't the number of API requests be limited by the number of plugins? 

Yes, let's not forget that we have more than 150 plugins and our limit is 60 per hour.

> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the user making requests.

https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting

> Using one of our own tokens is not great. If that person were to leave the organization, stuff would break immediately.

You're right then let's use the @bastianallgeier personal token 😅

lukasbestle:
Oh, 60 is indeed quite low. Then we have no other choice.

lukasbestle:
We will need a bot anyway for an automated release process, so I think it makes most sense to go that route.

afbora:
> We will need a bot anyway for an automated release process, so I think it makes most sense to go that route.

What exactly is it for? To periodically check the plugins (on repository) and update their versions?

lukasbestle:
No, not for the site, but we had the rough idea of automating Kirby releases across all our repos using GitHub Actions.

afbora:
I have no idea how to do it 🙈 Other than that, PR is ready.

lukasbestle:
We need Bastian to set up the email address (I pinged him in the headquarter). I will set that up and then pick up this PR from here. :)

afbora:
> I will set that up and then pick up this PR from here.

Thank you 👍 

> Also, the cache needs to be reset in the plugins presentation. Because all versions are false: https://getkirby.com/plugins.json

@lukasbestle What do you think about that? Should I create a new issue?

lukasbestle:
Is this just happening on the server because of the rate limit of 60 requests per hour? If so, the issue will solve itself on the next deployment once this PR is merged with the token authentication.

afbora:
When `/plugins.json` is called, it caches all json data and even if a new version of a plugin is cached, `plugins.json` data remains in the old.

As a solution, when a new version of a plugin is cached, it can trigger the `plugins.json` page to delete its cache.

Another solution is to periodically (1 or 3 hours) delete `plugins.json`'s cache.

lukasbestle:
I prefer the first solution to delete the cache on every update to any plugin cache. If you have time, you can add the change to this PR so we have it ready when we have that GitHub bot user token. Before that it doesn't really make much sense anyway because the plugin data is currently useless with the low rate limit.

afbora:
I've pushed a commit that triggers removing `plugins.json` cache when any plugin cache updates.

lukasbestle:
Let's see if that works in production. :)

afbora:
I've done a few tries, it works great 🎉

PR #1347, opened 2021-05-07
Add `--depth 1` to Docker examples

omBratteng:
Ideally the starterkit would've been cloned in a separate stage, and then copied to runtime stage. But as a starting example it's fine.
Adding the `--depth 1` to the command does slim it a bit down, and speeds up building the docker image


lukasbestle (approved):

texnixe (approved):

texnixe:
Makes total sense, thanks!

PR #1350, opened 2021-05-08
Add PHP tag to make example work

pReya:
* This example seems to be a complete listing. But it won't work without PHP opening tag.


lukasbestle:
Thanks! 👍

PR #1359, opened 2021-05-18
Add docs for table block, adapt template

texnixe:


afbora (approved):
It's good to see missing blocks for a long time 👍

distantnative:
Instead of the if statement for a specific page, I think we should use different field names for the two positions

distantnative:
@texnixe shall we change this still?

texnixe:
@distantnative Feel free, I don't quite understand what you mean with those different field names.

distantnative:
@texnixe I think it's bad practice if we start adding if-statements for one specific page in the templates, if the name changes or another page with the same needs comes up, this will get quite messy. So I thought, if you really want this text not in the standard position, we could add a `conclusion` field in addition to the `text` field which then gets output at the different position.

PR #1369, opened 2021-05-26
Add cookbook recipe for nested blocks

chrfickinger:


texnixe (approved):
The downloadable files probably also use short PHP tags (`<? `) which must not be used because they are likely to produce errors in environments which don't have them enabled. The short echo tags (`<?=`) are of course fine.

bastianallgeier:
I really love this. I have one minor concern though. We use camelCase for pretty much any example and you use underscores. Would you mind if we change that?

chrfickinger:
Absolutly not. I didn't know that, because this is my first contribution. Feel free to update the recipe as you want, i'll update the example files afterwards.

bastianallgeier:
I merged this manually.

PR #1373, opened 2021-05-28
Correcting type declaration in sample code

sebastiangreger:
As per advice from @distantnative [on Discord](https://discord.com/channels/525634039965679616/809460677658804234/847777733374377994)


lukasbestle:
I've switched this to an import with `use`, which makes this a bit cleaner. Thanks for your contribution! 👍

lukasbestle:
@distantnative I just noticed the param is documented as `Cms\Collection`, that's also what the native component receives. I wonder why it doesn't work with just `Collection` as that's aliased to `Cms\Collection`.

PR #1374, opened 2021-05-28
Sample code: replace pages search only

sebastiangreger:
As replacing `->search()` globally has a wide impact, the more likely use case for most developers might be to only customize the search for a specific collection type?

Probably obvious for experienced PHP developers, this example gives a copy-paste solution for only replacing one of them and fall back on the built-in defaults for the rest.

Feel free to edit or discard :)


lukasbestle:
Thank you! ❤️
I've used the opportunity to fix the code indentation in the existing code example and made some other minor improvements. The example is great and really useful.

PR #1378, opened 2021-06-02
Update email form cookbook recipe with the URL type for the honeypot field

renestalder:
The honeypot field in the cookbook recipe for an email form showed a website field using input type "website"
which doesn't exist in this form in the HTML specification. Presumly, what was meant is a field with the type "url", except the non-existing "website" type is some form of black magic tricking spambots into filling the field which I don't know of and that isn't explained in the example.

The type used in the example is now "url".


texnixe:
Thanks 🙈

PR #1380, opened 2021-06-04
Fix little error in code example for custom tags

darmstaedter:
Without the use statement the code is currently not working.


texnixe:
Thank you!

PR #1381, opened 2021-06-05
Create plugin-developer.txt

Werbschaft:
Tried to submit a plugin but don't think I did it right

### DIFF
--- a/content/plugins/werbschaft/plugin-developer.txt
+++ b/content/plugins/werbschaft/plugin-developer.txt
@@ -1,0 +1,1 @@
+Title: Werbschaft


lukasbestle:
As far as I can tell, the plugin itself wasn't yet merged, right? Please add the plugin directory to the PR as well to avoid an "empty" author directory.

Werbschaft:
Should be done, but am not sure if everything is correct. Maybe the method of submitting plugins should be reconsidered, it's really not very easy, especially for beginners.

Edit: despite resubmission it seems not to have worked. Maybe it has to do with the fact that the creation of the subfolder for the plugin creator was not yet approved.

lukasbestle:
Thanks, I've merged both PRs. 👍

It looks like you have created the PRs from the GitHub web UI, right? By default it only allows changes to one file at a time. You can manually edit and create additional files in the automatically created patch branch before you create the pull request. Combining all changes in one pull request makes it easier to review the changes and also avoids incomplete merges.

An alternative and the recommended way is to clone your fork locally and make the changes there. This also enables you to preview how the plugin entry will look like on the site.

I agree the current submission process is not as easy as it could be. Yesterday we actually thought about creating a submission form, but it would be complex to implement correctly. In any case we have plans to extend and improve the plugin directory.

Werbschaft:
Yes I did it via the web interface. I also noticed that I can only create one file, sorry for the mess. As I said, I'm not very familiar with GitHub. A simpler submission process would be great, but I can well imagine that it is very complex to implement such a thing in the current process

PR #1383, opened 2021-06-06
Add virtual files recipe

texnixe:


lukasbestle (approved):
Awesome! 👍😍
I've committed some fixes and improvements directly, I hope that's OK.

texnixe:
Of course, @lukasbestle, thanks for checking it out!

PR #1384, opened 2021-06-07
Update description for File::setParent()

### LINKED ISSUE #1348: Required props in the references are missing
As already mentioned at Discord, some methods lack the required props. An example is available at:

https://getkirby.com/docs/reference/objects/cms/file/create#parameters

Here the `source` is missing in the documentation, without the method does not work at all. As I know `parent` is also required, but not highlighted as it. @distantnative meant that this still some other sides concerns.

lukasbestle:
Fixes #1348.

Generally we don't update the `kirby` dir in getkirby.com manually, but I think here it makes sense to already document this now and not wait for the 3.6.0 release. I have also pushed the change to the PR branch for https://github.com/getkirby/kirby/pull/3372.

### DIFF
--- a/kirby/src/Cms/File.php
+++ b/kirby/src/Cms/File.php
@@ -688,7 +688,9 @@ class File extends ModelWithContent
     }
 
     /**
-     * Sets the parent model object
+     * Sets the parent model object;
+     * this property is required for `File::create()` and
+     * will be generally required starting with Kirby 3.7.0
      *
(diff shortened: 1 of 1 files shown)


distantnative:
I remember @bastianallgeier used to be quite opposed to using anything than actual releases, however I'd agree with you for all the reasons you've outlined and that it only documents already existing behavior

PR #1386, opened 2021-06-08
Bugfix: Lonely pipe character (vertical bar) is not shown in code

Adspectus:
On the Kirby website in cookbook, recipe "Find translations in your code" the source code contains (\`|`) to show a vertical bar in code style. However, the lonely vertical bar is not shown at all. Thus, I had to change the piece of code here.


distantnative:
@texnixe are you fine with this way to fix it? We could also fix it in theist's code - I am pretty sure it originates from all in-line codes being checked for types (string, float, objects) and there the separator is considered as splitting element between two types and thus disregarded. We could add an if statement for single `|` to be exempt there, I think.

texnixe:
I don't know who `theist` (I guess prism) is, but I think we should change it in the code, as this problem might come up in other contexts as well and we should mark code as code and not in quotes.

distantnative:
Ups, sorry "the site's"

lukasbestle:
I agree with Sonja.

Adspectus:
Ok, but what does it mean for the bug in the text of the recipe?

distantnative:
Fixed it directly https://github.com/getkirby/getkirby.com/commit/fb642498e09fee700777ff61950ee38d218c83b6

Thanks for pointing this out and helping to fix it!

PR #1387, opened 2021-06-08
Ignore editors

afbora:


lukasbestle:
If we do this, we also need to explicitly delete the `.vscode` directory in a commit, otherwise the `.gitignore` won't have an effect on it.

afbora:
How can I do that? Because I'm using PhpStorm.

lukasbestle:
You can just delete the `.vscode` folder in your file manager. Git will then see the change and offer to commit the deletion to the PR.

afbora:
I didn't realize it had the `.vscode` folder because I was using PHPStorm 😅

lukasbestle (approved):

afbora:
👉  Would it be okay for you to ignore IDE files/folders? @bastianallgeier @distantnative @texnixe

bastianallgeier (approved):

bastianallgeier:
Yep, totally fine for me.

afbora:
🎉 Thank you

PR #1389, opened 2021-06-08
New submit plugin form

afbora:


distantnative:
@afbora I forgot it already - maybe it's good to keep the status in the first comment updated:
What was the status regarding the GitHub API? Is there a way without private keys?

PR #1391, opened 2021-06-09
Remove redundant line from submodule example

keybits:
`git add kirby` causes an error in this example. The command is not needed is not needed to successfully achieve the recipe.


texnixe:
What error do you get? I cannot add the submodule without staging the removed `kirby` folder first.

keybits:
@texnixe Thanks for the reply. I just realised what I did. I didn't follow the instructions in order and had downloaded the starterkit rather than cloning it. I then deleted the Kirby folder before doing a `git init`, so Git never knew about the Kirby folder. Sorry for the noise!

PR #1401, opened 2021-06-22
Missing &

dinko:
According to an ngnix admin I'm working together for a project, there's a missing ampersand…please check.


bastianallgeier (commented):
I cannot confirm this. We use the same setup for our site and the ampersand is not needed. @pReya what do you think?

dinko:
Ooops…the admin just confirmed the ampersand is actually not necessary. Never mind — sorry for the confusion.

PR #1404, opened 2021-06-24
New cookbook recipe: nested blocks advanced

chrfickinger:


content/docs/2_cookbook/10_panel/0_nested-blocks-advanced/cookbook-recipe.txt
```diff
@@ -0,0 +1,99 @@
...
+
+### FAQ snippet
+We can deliver different HTML code depending on which version we are in by checking the `layout` value.
+
+```php
+$layout = $block->layout()->value();
+
+if ($layout == 'v1'):
+
+  // HTML code for layout version 1
```
distantnative, line 61: Would be great for beginners to have actual HTML code examples here, helps them often to not get lost if we keep a more concrete than abstract.

content/docs/2_cookbook/10_panel/0_nested-blocks-advanced/cookbook-recipe.txt
```diff
@@ -0,0 +1,99 @@
...
+
+  snippet('blocks/' . $nestedBlock->type(),[
+    'block' => $nestedBlock,
+    'layout' => $layout
+  ]);
+
+endforeach;
+```
+
+
```
distantnative, line 82: I know when writing recipes it might seem redundant and totally not DRY as we used to, but I think it helps to show at the end of such a section how the code looks together. So readers don't need to wonder "so where does the foreach loop go", but can see it directly.

content/docs/2_cookbook/10_panel/0_nested-blocks-advanced/cookbook-recipe.txt
```diff
@@ -0,0 +1,99 @@
...
+### Question/answer snippet
+
+By passing the layout value from the parent block, we can use it via `$data['layout']` in the nested blocks now. We just need to check the value and can deliver different HTML output for each version.
+
+```php "/site/plugins/faq-block/snippets/blocks/faqItem.php"
+if ($data['layout'] == 'v1'):
+
+  // Nested HTML code for layout version 1
```
distantnative, line 92: Same here, actual HTML code examples would be great.

distantnative (commented):
@chrfickinger Thanks for this - it's really cool!
I would suggest to make it a bit more concrete in places (see comments) - that helps beginners a lot to not get lost in abstraction.

chrfickinger:
You are right! Actually, it's just an extension of my recipe for nested blocks. It is definitely necessary to install the mentioned recipe (which has a lot of context) to change with the new functionality. 

Do you think I should add this recipe as a new paragraph in the nested blocks recipe?

chrfickinger:
@texnixe what do you think? 2 recipes or 1 ?

texnixe:
@chrfickinger How about using a simpler example, like one of the standard block types as a basis, and then extend them with a layout field? That way, we had a standalone version, that doesn't depend on another recipe, and has more universal value?

In any case, I think it would be important to add the HTML (and maybe even some basic styling, but that's secondary) so that readers can really follow along. For beginners, it would otherwise probably be a bit too abstract.

PR #1412, opened 2021-07-05
Fixed typo

Jeanettt:
Switched around hasNextPage(), hasPrevPage(), prevPageUrl() and nextPageUrl() in code example.

### DIFF
--- a/content/docs/2_cookbook/3_templating/0_pagination/cookbook-recipe.txt
+++ b/content/docs/2_cookbook/3_templating/0_pagination/cookbook-recipe.txt
@@ -43,14 +43,14 @@ Add this below your foreach loop:
 <?php if ($articles->pagination()->hasPages()): ?>
 <nav class="pagination">
 
-  <?php if ($articles->pagination()->hasNextPage()): ?>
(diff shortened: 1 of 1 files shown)


distantnative:
Are you sure? I think the code is correct already:
We flip the pages collection, so in `$articles ` the newest articles are at the front, the oldest at the back. So for newer articles you need to check the previous one, for older articles the next.

Jeanettt:
Woops sorry I guess you're right! I was using it in the opposite way and therefore thought it was upside down :)

Have a good day!
________________________________
Van: Nico Hoffmann ***@***.***>
Verzonden: maandag 5 juli 2021 16:38
Aan: getkirby/getkirby.com ***@***.***>
CC: Jeanettt ***@***.***>; Author ***@***.***>
Onderwerp: Re: [getkirby/getkirby.com] Fixed typo (#1412)


Are you sure? I think the code is correct already:
We flip the pages collection, so in $articles the newest articles are at the front, the oldest at the back. So for newer articles you need to check the previous one, for older articles the next.

afbora:
No problem. Thanks for your contribution though 👍

PR #1417, opened 2021-07-16
Update reference-article.txt

rlb222:
I couldn't find the recommendation to remove the setting after installation anywhere in the docs, so I added it here.


texnixe:
Thanks!

PR #1418, opened 2021-07-16
Designers page: Add section on white-labeling

### LINKED ISSUE #1414: Add white-label Panel interface as feature
That our Panel is basically a white-label solution is a big feature that often comes at extra costs with other systems. We don't sell it on our site at all.

lukasbestle:
Closes #1414.

**Open question:** As far as I can tell, it's currently not possible to set a custom favicon for the Panel. This would be the missing piece for a fully white-label solution.

### DIFF
--- a/content/features/2_designers/features-for-designers.txt
+++ b/content/features/2_designers/features-for-designers.txt
@@ -9,6 +9,7 @@ Benefits:
 - Design without compromises
 - Clear separation between design and content
 - A fully customizable admin interface
+- White-labeled from the start
 - A fantastic prototyping companion
 
 ----
--- a/site/snippets/templates/features-for-designers/workflow.php
+++ b/site/snippets/templates/features-for-designers/workflow.php
@@ -20,7 +20,7 @@
     ],
     [
       'title' => 'An interface without branding',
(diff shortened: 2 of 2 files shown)


bastianallgeier:
I would still remove the favicon sentence and just get that fixed in 3.6 with a panel.favicon config option.

lukasbestle:
I think we should mention the favicon. We can't really say "there isn't any branding whatsoever by default".

👍 on the `panel.favicon` option for 3.6: https://github.com/getkirby/kirby/issues/3516

distantnative:
I'd remove that favicon sentence part. It reads weird in there - moving from the big inspirational picture to a tiny detail. I think there are a lot of solutions out there as "white-labeled" that ship with some defaults that easily can be deactivated. Our only thing is the favicon which will be deactivatable in 3.6 - I think nobody except us would blame us for stating "there is no branding".

lukasbestle:
You are right, other companies would be even more aggressive on their marketing pages. I rephrased it a bit. :)

distantnative (approved):

bastianallgeier (approved):

PR #1431, opened 2021-07-26
Add new cookbook recipe different blueprints - same template

texnixe:


bastianallgeier (approved):
This is a really nice idea!

PR #1452, opened 2021-08-07
Fix return for `apply` hooks

### LINKED ISSUE #1372: Apply hooks return multiple variables
Example: http://getkirby.com/docs/reference/plugins/hooks/kirbytags-before

distantnative:
Fixes #1372 

@texnixe does this work for you too?


texnixe:
Yep

PR #1454, opened 2021-08-08
Corrections 3.6 release page

afbora:


distantnative:
We should mention `<k-outside>` as well as another option

afbora:
Since I don't know `<k-outside>` fully, I didn't write it. Because when I don't use any inside or outside, it looks like `<k-outside>`. What is the difference using `<k-outside>`  or without `<k-outside>`?

afbora:
Also, if the `inside` and `outside` functions are for fully related authorized users, what would a clearer naming be? Like `k-auth` and `k-guest` or something like that 🤔

distantnative:
I think inside and outside already fit quite well as name. The one is for when a user is inside the panel (signed in), the other for all that can also be accessed when signed out

afbora:
Got it, so can you add the explanation about outside? @distantnative

afbora:
Also what is the difference? Because both views are same.

````vue
<template>
  <k-header>Todos</k-header>
</template>
````
and
````vue
<template>
  <k-outside>
    <k-header>Todos</k-header>
  </k-outside>
</template>
````

**Comment by distantnative** [comment]
Not much, adds this wrapper: https://github.com/getkirby/kirby/blob/develop/panel/src/components/Layout/Outside.vue

**Comment by afbora** [comment]
I added the `<k-outside>` description with example. If it needs a correction, I'd be happy if you could take care of it.

File: content/releases/3-6/5_breaking-changes/release-guide.txt [hunk 1]
```diff
@@ -55,8 +55,27 @@ We know that breaking changes suck, when they break your plugins or other parts
 - Direct URLs for the users view have changed, e.g. now `/panel/users/?role=yourRole`
 
 ### Custom Panel views
-- Using custom Panel views (incl. routing, breadcrumbs...) in plugins has changed fundamentally. You will have to use the [new Panel areas feature](/releases/3.6/features/fiber#area-definition) now.
+- Using custom Panel views (incl. routing, breadcrumbs...) in plugins has changed fundamentally. You will have to use the [new Panel areas feature](/releases/3.6/fiber/areas#area-definition) now.
 - Accordingly, the frontend extension `panel.plugin({ view: [...] })` has been  removed. Use Panel areas instead.
+- If the custom panel view is for authenticated users, you will have to wrap your custom view with `<k-inside>`. If not, you should wrap it `<k-outside>`.. [See more component changes](/releases/3.6/breaking-changes#panel__vue-component-changes)
```
texnixe: panel => Panel
you should wrap it `<k-outside>`.. => you should wrap it with `<k-outside>`.

content/releases/3-6/5_breaking-changes/release-guide.txt
```diff
@@ -55,8 +55,27 @@ We know that breaking changes suck, when they break your plugins or other parts
...
+
+```html "For authenticated page"
+<template>
+  <k-inside>
+    ...
+  </k-inside>
+</template>
+```
+
+```html "After signed out for guests"
```
texnixe: After signed out for guests => After logout for visitors

texnixe (commented):
See comments

content/releases/3-6/5_breaking-changes/release-guide.txt
texnixe: Should have read the thread first...

afbora: For unauthorized users. I guess it could be for a public page using the panel. I wrote it according to what @distantnative said 🙈

content/releases/3-6/5_breaking-changes/release-guide.txt
afbora: Thanks, done.

content/releases/3-6/5_breaking-changes/release-guide.txt
texnixe: Yes, I deleted my original comment, because I missed the thread and only looked at the changes, sorry.

afbora: Ok, done.

bastianallgeier (approved):

PR #1463, opened 2021-08-21
Append permission pitfall

scsskid:


distantnative:
Instead of setting permissions for the whole Kirby folder, wouldn't it bet better to create the media folder and set permissions for it?

scsskid:
I guess you are right

maybe link to existing docs where is stated which folders are written to?

site/cache
site/sessions

...

?

distantnative:
@texnixe do we have those listed somewhere? I can't find anything.

texnixe:
@distantnative Doesn't look like we have such a list in the docs, although the Panel checks at installation if these folders are writable (although the cache folder is not part of the check, no idea why).

scsskid:
I narrowed it down to:

/media
/content
/site/sessions
/site/accounts
/site/cache

texnixe:
@scsskid Yes, those are the ones. That these folders are writable, is a general requirement though, not limited to the nginx cookbook recipe.

scsskid:
fair enough, closing

PR #1464, opened 2021-08-24
nest level of the tailwind.css in the src folder

igorlanko:
Currently, the folder structure figure shows that `tailwind.css` is placed alongside the `css` folder, while it says "we create a subfolder `/css` in the `/src` folder and inside that folder a file called `tailwind.css`.".


distantnative:
@chrfickinger would be great to get your view on this.

PR #1476, opened 2021-09-08
Add cookbook recipe: Batch update content

sebastiangreger:
As this is a task that keeps popping up in the forum etc., I finally got around to write up my approach for updating large amounts of content (I've done this several times with up to 10.000 pages per run, and refined this solution slowly over time).

Written to build up gradually, from explaining the underlying logic all the way to dealing with server time-outs on large batches, I tried to keep it as barebone as possible (added in a few examples though, to illustrate how to deal with different types of content). Code snippets have been tested with a fresh Starterkit setup (Kirby 3.5).

I'm still a bit unsure should I add in more checks (for example, the current code fails if the field to update doesn't actually exist on some page in the collection looped over), which would make it more robust for copy-paste reuse, but also a tad more complex. Feedback welcome; I could add those validation checks in, no problem. Better to keep it short or add extra code that makes the experience of using the sample code more smooth?

### DIFF
--- a/content/docs/2_cookbook/2_content/0_batch-update/cookbook-recipe.txt
+++ b/content/docs/2_cookbook/2_content/0_batch-update/cookbook-recipe.txt
@@ -94,7 +94,11 @@ return [
 ];
 ```
 
-Now our update loop can be triggered using the URL `mydomain.com/batch-update-content`. This includes a check to verify admin status; the route will fail unless logged in. It is a good idea to delete or comment-out this route when done, to keep things clean and safe.
+Now our update loop can be triggered using the URL `mydomain.com/batch-update-content`. This includes a check to verify admin status; the route will fail unless logged in.
+
+<warning>
+It is a good idea to delete or comment-out this route when done, to keep things clean and safe.
+</warning>
 
 On multi-language sites, we have to specify a language both on read and write. `$page->content()->myfield()->value()` becomes `$page->content('en')->myfield()->value()` to retrieve the desired content, and `$page->update(['myfield' => $newValue])` receives an additional second argument with the target language: `$page->update(['myfield' => $newValue], 'en')`.
 
@@ -194,7 +198,7 @@ foreach($collection as $page) {
 }
 ```
 
-Similarly, we could change other attributes of Blocks objects; for instance we could manipulate the `$old['type']` variable to make turn all blocks of type `text` into a (hypothetical) custom block type `mytext`, etc.
+Similarly, we could change other attributes of Blocks objects; for instance we could manipulate the `$old['type']` variable to convert all blocks of type `text` into a (hypothetical) custom block type `mytext`, etc.
 
 ### Updating files
 
--- a/content/docs/2_cookbook/2_content/0_batch-update/cookbook-recipe.txt
+++ b/content/docs/2_cookbook/2_content/0_batch-update/cookbook-recipe.txt
@@ -1,0 +1,349 @@
+Title: Batch updating content
+
+----
+
+Published: 2021-08-30
+
+----
+
+Description: Flexible and reliable updates to large amounts of content
+
+----
+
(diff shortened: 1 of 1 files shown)


texnixe:
Hey @sebastiangreger Great recipe! I think we can leave it as is without the additional checks but should then add a note that this is not a copy&go recipe.

Not a must, but since this comes up ever so often, it would be nice to have an example for multi-language sites. What do you think?

sebastiangreger:
Thank you for your feedback, @texnixe 

> I think we can leave it as is without the additional checks but should then add a note that this is not a copy&go recipe.

Good to hear you agree; I, too, believe it is better to keep it simple …makes it is easier to follow along. I added a note about the limitations.

> Not a must, but since this comes up ever so often, it would be nice to have an example for multi-language sites. What do you think?

Very important point that I completely forgot about; language versions absolutely have to be addressed in the recipe! I decided against adding more examples, as (if I'm not overlooking anything) it is really just about adding the language code to read/write operations. So I decided to globally switch from the `$page->myfield()` shorthand to the formally correct `$page->content()->myfield()` in all examples (also reduces potential for error if field names collide with page methods' names) and added a paragraph about adding the language code to the `content()` method as well as to the `update()` call.

I also did a few minor edits for style and fixed some typos.

Please feel free to edit the sample code or any text if my style is not 100% somewhere or to ask for further improvements if you still see something missing :slightly_smiling_face:

content/docs/2_cookbook/2_content/0_batch-update/cookbook-recipe.txt
```diff
@@ -0,0 +1,392 @@
...
+
+                    }
+                }
+            }
+        ],
+    ]
+];
+```
+
+Now our update loop can be triggered using the URL `mydomain.com/batch-update-content`. This includes a check to verify admin status; the route will fail unless logged in. It is a good idea to delete or comment-out this route when done, to keep things clean and safe.
```
texnixe: I'd move the last sentence into a warning block

content/docs/2_cookbook/2_content/0_batch-update/cookbook-recipe.txt
```diff
@@ -0,0 +1,392 @@
...
+
+    // update the content file
+    $page->update([
+        'text' => $blocks
+    ]);
+    
+}
+```
+
+Similarly, we could change other attributes of Blocks objects; for instance we could manipulate the `$old['type']` variable to make turn all blocks of type `text` into a (hypothetical) custom block type `mytext`, etc.
```
texnixe: make turn => turn or convert

texnixe (changes requested):
I found only two small issues.

As already discussed on the forum a while ago, another option for large data sets would be to run a script on the command line. But we can move this to a future recipe which is on my list anyway and reference that later.

sebastiangreger:
> I found only two small issues.

Thanks, I fixed the sentence and added a warning block (I don't currently have a setup to render the recipe into its final layout, so tried to be a bit cautious on using too many blocks as I can't picture how this whole thing will look like when rendered; but this indeed should be highlighted as a warning).

> As already discussed on the forum a while ago, another option for large data sets would be to run a script on the command line. But we can move this to a future recipe which is on my list anyway and reference that later.

The CLI is of course the straightforward way of doing it and pairing these two recipes sounds like a great idea to show the alternatives :+1: I altered the sentence about the need for a staggered process to mention the CLI already, as I think it is important to highlight that this is an alternative, not the only solution.

To briefly illustrate my motivation for this "workaround approach": I hoped to find a robust and reusable solution that does not require shell access (one of the great benefits of Kirby being that it can be run on the most minimal hosting packages) and avoids some headaches with PHP on shell (e.g. when the web server is configured to use a different version than the SSH user) and hence has a low barrier for entry.

texnixe (approved):

PR #1479, opened 2021-09-12
Update loops cookbook recipe

matthiasott:
Minor correction: The index of nth() starts at zero, so the selected element in the example is actually the fourth.


texnixe:
Thanks for correcting!

Issue #1434, opened 2021-07-30
parcel-bundler (v1) ist not maintained anymore


arnoson:
https://getkirby.com/docs/guide/plugins/plugin-setup-panel#introducing-parcel uses `parcel-bundler` which for has stopped working for me throwing various errors and is not maintained anymore. I tried to build a kirby plugin with parcel 2 but didn't have any success either. Would be great if the plugin-setup-panel guide could be updated to parcel 2 or an alternative bundler.

distantnative:
I don't think this I quite correct. Parcel 2 is still not out of beta/RCs either.

texnixe:
I wonder if we should use a different bundler, I'm also running into issues with the current setup time and again. Problably related to node.js versions.

lukasbestle:
I also had several issues with Parcel 1, but couldn't get Parcel 2 to work either. :(
Unfortunately I don't have experience with other "modern" bundlers. In the end I could get Parcel 1 to work somehow, so I didn't investigate further.

texnixe:
Yes, I always got Parcel 1 to work again as well, but always a bit of a pain.

@rasteiner posted his Rollup solution on the forum recently: https://forum.getkirby.com/t/build-failed-on-parcel/22912/6

But I guess it's always the same with those bundlers, you change your environment and the stuff stops working. Guess I'll try a Docker approach in the future.

arnoson:
Thanks @texnixe for posting the link to the rollup setup!
It is true that bundlers always complicate things and at some point might stop working, but I have the feeling with parcel it is particularly bad. As @distantnative mentioned parcel 2 is still beta, but parcel 1 is already not maintained anymore which seems weird. I think going for rollup would be a better choice, since the development on it has been running reliably for a long time.
The Vue ecosystem also relies on rollup (vite) so it seems to be a good choice in terms of vue as well.

rasteiner:
Since the panel has switched to vite, maybe it's possible to get it to work also for plugins. The rollup script I posted on the forum doesn't do HMR, so maybe that could be done better with vite. I don't have much experience with it though...

renestalder:
If somebody is interested in a esbuild solution, here is a version that works for me:

`package.json`:
```json
{
  "scripts": {
    "serve": "node scripts/watch",
    "build": "node scripts/build"
  },
  "dependencies": {
    "vue": "^2.6.14",
    "vue-hot-reload-api": "^2.3.4",
[...]

PR #1490, opened 2021-09-22
Replace Parcel with kirbyup

johannschopplich:
It's finally here! Parcel v1 can now rest in peace in favor of [kirbyup](https://github.com/johannschopplich/kirbyup). I have changed all occurrences (excluding Kirby Kosmos issues) and usage instructions.

We can even drop the global (or otherwise local) install instructions, since we simply use `npx` to run kirybup. Thus, no `devDependencies` inside the `package.json`. It will be cached globally and reusable for other Panel plugins, but the user won't have to do a thing. From:

```json
{
  "scripts": {
    "dev": "parcel watch src/index.js --no-source-maps -d ./",
    "build": "parcel build src/index.js --no-source-maps --experimental-scope-hoisting -d ./"
  },
  "posthtml": {
    "recognizeSelfClosing": true
  }
}
```

To:

```json
{
  "scripts": {
    "dev": "npx -y kirbyup src/index.js --watch",
    "build": "npx -y kirbyup src/index.js"
  }
}
```

Thus, no need to run either `npm i`, nor `npm i -g kirbyup`!

I hope the changes appeal to you. Feedback welcome. And: Happy 3.6 beta release! What do you think about mentioning kirbyup in the release changelog? Some developers may want to switch.

Related: https://github.com/getkirby/getkirby.com/issues/1434

### DIFF
--- a/content/docs/1_guide/17_plugins/6_plugin-setup-panel/guide.txt
+++ b/content/docs/1_guide/17_plugins/6_plugin-setup-panel/guide.txt
@@ -14,37 +14,27 @@ Our Pluginkit for Panel plugins is almost identical to the basic setup – excep
 
 ## Installation
 
-We have created a simple example plugin called the Pluginkit.
-You can find the code for this tutorial in the (link: https://github.com/getkirby/pluginkit/tree/4-panel text: `4-panel`) branch.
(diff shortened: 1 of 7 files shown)


bastianallgeier (approved):
You are a hero!

PR #1501, opened 2021-10-03
Add template: to pages section options

rlb222:
(section-options: pages)
In this list I want to change 'templates' into 'template: or templates: '.
I had been searching a long time to find this, and I couldn't find "option template:" anywhere. Until I followed the link to panel/sections/pages  _and_ scrolled down beyond the options list.    
(I know I can't change the options myself, I believe they are generated somehow)

### DIFF
--- a/content/docs/3_reference/3_panel/4_sections/0_pages/reference-panelsection.txt
+++ b/content/docs/3_reference/3_panel/4_sections/0_pages/reference-panelsection.txt
@@ -71,7 +71,7 @@ Keep in mind that the same section name can only be used once per blueprint.
 
 The pages section has multiple options to control what kind of pages should be displayed, how they should be displayed and what happens if a new page is added.
 
-(section-options: pages)
+(section-options: pages) 
 
 ## Headline
 


texnixe:
Did you create this PR by accident? I don't really understand the purpose of the space...

rlb222:
I wanted to suggest to add ‘template:’, but the options are not editable (at least I couldn’t find it).
So the only thing I  could do was add a space and put the explanation in the 

I’m sorry if that was not the right way to do it.

René 

This is the explanation I wrote:

(section-options: pages)
In this list I want to change 'templates' into 'template: or templates: '.
I had been searching a long time to find this, and I couldn't find "option template:" anywhere. Until I followed the link to panel/sections/pages  and scrolled down beyond the options list.
(I know I can't change the options myself, I believe they are generated somehow)

---//---

texnixe:
Closing this because adding a space after the tag doesn't solve the problem. Most of the documentation is pulled in automatically and cannot be changed. The property is actually called `templates` not `template`, therefore, there is no entry in the table, but `template` can be used as well for single templates (but the plural works as well for single templates).

PR #1503, opened 2021-10-04
fix srcset code example

tobystokes:
A minor change, and it seems it is correct in the official documentation, but took me a while to spot this one, and you make it so easy to PR...

(And the results of my testing this feature on `3.6.0-beta.2`: works a treat!)


texnixe:
Thanks!

PR #1420, opened 2021-07-17
Docs for 3.6.0

distantnative:
Project with todos: https://github.com/orgs/getkirby/projects/5/views/1

PR #1512, opened 2021-10-21
Add `email` to the available writer marks

timnarr:
Small doc improvement... Added `email` as a available format to the writer marks.

### DIFF
--- a/content/docs/3_reference/3_panel/3_fields/0_writer/reference-article.txt
+++ b/content/docs/3_reference/3_panel/3_fields/0_writer/reference-article.txt
@@ -54,6 +54,7 @@ The writer field supports the following marks by default:
 - `strike`
 - `code`
 - `link`
+- `email`
 
 ```yaml
 fields:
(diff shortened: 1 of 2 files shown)


distantnative:
@afbora does this refer to this change? https://getkirby.com/releases/3.6/features/email-writer-dialog
Then it would only be in 3.6

Also @TimGr we don't edit the core files (everything in `kirby/` folder) in this repo as we use a standard kirby installation to run the site. We would need to make that change in https://github.com/getkirby/kirby

timnarr:
Ah, I see thanks for the hint @distantnative 
I'm currently working on a site that runs `3.6.0-rc.1` and that's where I noticed this. So maybe you're right that this refer to the mentioned change

afbora:
@distantnative Yes, I've created PR for it: https://github.com/getkirby/kirby/pull/3835

afbora:
Also I've added to 3.6 docs: https://github.com/getkirby/getkirby.com/pull/1420

bastianallgeier:
✅

PR #1517, opened 2021-10-24
Add note: toPages() only returns published pages

sebastiangreger:
via https://forum.getkirby.com/t/draft-page-title-not-accessible-in-pages-method/23762


texnixe:
Thanks, @sebastiangreger!

PR #1518, opened 2021-10-24
Update Kirby Annotator plugin to its current name

trych:
The Kirby Annotator plugin seems to have been renamed from Kirby Image Annotator to its current name Kirby Annotator. This edit makes sure the correct URL is used for the `plugins.json` route and others.


lukasbestle:
Thanks!

PR #1530, opened 2021-11-01
Texnixe/3.6.0

texnixe:


content/docs/3_reference/7_plugins/1_extensions/0_panel-login/reference-extension.txt
```diff
@@ -1,11 +1,16 @@
 Title:  Panel login
 
 ----
+Text:
 
-Since: 3.2.0
+<since v="3.6.0">
+<warning>
+Panel views have been replaced with (link: docs/reference/plugins/extensions/panel-areas text: Panel areas) in 3.6.0.
+</info>
```
distantnative: I think this `</info>` shouldn't be there

content/docs/3_reference/7_plugins/1_extensions/0_panel-views/reference-extension.txt
```diff
@@ -4,6 +4,13 @@ Title:  Panel views
 
 Text:
 
+<since v="3.6.0">
+<warning>
+Panel views have been replaced with (link: docs/reference/plugins/extensions/panel-areas text: Panel areas) in 3.6.0.
+</info>
```
distantnative: Same here

content/docs/1_guide/7_languages/1_introduction/guide.txt
```diff
@@ -76,16 +76,22 @@ If you don't use the Panel, you have to add the language extension to your text
 
 ### Via the Panel
 
-After enabling languages in your config as described above, you can find the language setup on the Panel in the "Settings" page.
+In versions prior to 3.6.0, you can find the language setup in the "Settings" page. Managing languages works in the same way.
```
distantnative: Based our discussion of today - creating separate docs copies e.g. for 3.5 - I would suggest that we remove anything about pre-3.6 versions here.

content/docs/3_reference/3_panel/1_blueprints/0_page/reference-article.txt
```diff
@@ -136,6 +136,54 @@ icon: page
 icon: 📚
 ```
 
+## Image options
```
distantnative, line 139: Let's create this as snippet, e.g. `snippets/docs/panel-field-image.php`, and then reference it in the various reference pages instead of having the same text there as copies. What do you think?

content/docs/3_reference/3_panel/4_sections/0_pages/reference-panelsection.txt
```diff
@@ -123,6 +123,21 @@ layout: cards
 
 (image: cards.png)
 
+
+### Cardlets layout
+
```
distantnative, line 128: Same here, I would add it as snippet and reference it instead the same text in multiple pages

content/docs/3_reference/7_plugins/1_extensions/0_panel-login/reference-extension.txt
```diff
@@ -1,11 +1,15 @@
...
 
 ----
+Text:
 
-Since: 3.2.0
+<since v="3.6.0">
+<warning>
+Panel views have been replaced with (link: docs/reference/plugins/extensions/panel-areas text: Panel areas) in 3.6.0.
+</warning>
+</since>
```
distantnative: We could use the `Deprecated`  field instead:

```
Deprecated:
3.6.0|Panel views have been replaced with (link: docs/reference/plugins/extensions/panel-areas text: Panel areas)

**Comment by distantnative** [comment]
Looks really good - great chunk of work. I added a few more comments.

File: content/docs/3_reference/3_panel/1_blueprints/0_page/reference-article.txt (hunk 4 above)
Comment (texnixe) [reply content/docs/3_reference/3_panel/1_blueprints/0_page/reference-article.txt:139]: I had originally intended to do that, but we would have to extend the docs tag to pass attributes, because the text is not exactly the same in all instances.

File: content/docs/3_reference/7_plugins/1_extensions/0_panel-login/reference-extension.txt (hunk 6 above)
Comment (texnixe) [reply content/docs/3_reference/7_plugins/1_extensions/0_panel-login/reference-extension.txt]: With the  shift to separate docs for 3.6, these pages should be deleted, anyway, shouldn't they?

Comment (distantnative) [reply content/docs/3_reference/7_plugins/1_extensions/0_panel-login/reference-extension.txt]: True 👍

File: content/docs/3_reference/3_panel/1_blueprints/0_page/reference-article.txt (hunk 4 above)
Comment (distantnative) [reply content/docs/3_reference/3_panel/1_blueprints/0_page/reference-article.txt:139]: Are there multiple differences? I only spotted one word - `pages`/`users`. If that would be the case, we can already do that. the `docs` tag already allows for a `field` attribute that we could use then.

Comment (texnixe) [reply content/docs/3_reference/3_panel/1_blueprints/0_page/reference-article.txt:139]: There are also the examples, e.g., including the path to the example files: (not visible here)


```yaml "site/blueprints/pages/note.yml"
image:
  back: "{{ page.myCustomBackColor }}"
```
vs

```yaml "site/blueprints/users/editor.yml"
image:
  back: "{{ user.myCustomBackColor }}"
```

distantnative: I missed that! Then let's leave it manual as it is right now.

distantnative (approved):

PR #1535, opened 2021-11-06
[3..6] Remove old `<since>` blocks

distantnative:


texnixe:
Looks good. I think we should also remove old since fields?

distantnative:
That's probably something to discuss. I left them as we will also pull in a lot of `@since` tags from the code directly. So I thought, maybe it's fine to have those little page-wide tags still.

texnixe (approved):

PR #1536, opened 2021-11-06
Cookbook recipe multilang field options

texnixe:


content/docs/2_cookbook/10_panel/0_multilang-field-options/cookbook-recipe.txt
```diff
@@ -0,0 +1,163 @@
...
+                        'photography'  => I18n::translate('photography', null, $lang),
+                        'design'       => I18n::translate('design', null, $lang)
+                    ],
+                    'de' => [
+                        'architecture' => I18n::translate('architecture', null, $lang),
+                        'photography'  => I18n::translate('photography', null, $lang),
+                        'design'       => I18n::translate('design', null, $lang)
+                    ],
+                ];
+                return $categories[$lang] ?? [];
```
lukasbestle: Aren't both language arrays the same now? I guess it can be rewritten to this:

```suggestion
                return [
                    'architecture' => I18n::translate('architecture', null, $lang),
                    'photography'  => I18n::translate('photography', null, $lang),
                    'design'       => I18n::translate('design', null, $lang)
                ];
```

lukasbestle (commented):
I like it! Somehow it has that "needs to be a core feature" smell as well, but the solution you explain in this recipe is pretty simple as well and also shows the options API feature and its flexibility for even more custom behavior.

content/docs/2_cookbook/10_panel/0_multilang-field-options/cookbook-recipe.txt
texnixe: You are of course right 🙈 ! And I'm even too stupid to just commit the suggestion 🙈 🙈 🙈 .

texnixe: > needs to be a core feature

Not sure about this, because you  end up with mixed languages in the form.

lukasbestle (approved):

content/docs/2_cookbook/10_panel/0_multilang-field-options/cookbook-recipe.txt
lukasbestle: Yes, it's difficult to do in the current blueprint syntax, but having to use the options API is not ideal either.

texnixe: It just dawned on me that this use case can easily be achieved with dynamic blueprints. I think we can throw this recipe away.

texnixe: And I will add this use case to the other recipe instead.

PR #1553, opened 2021-11-15
Enhancements to release page

lukasbestle:
Here's the first round of my suggestions for the release page.
I didn't find the time to look at the updated release notes yet. Will do so tomorrow morning.

What I noticed is that the tables in the Core Refactoring and Deprecated sections are too narrow for the many columns. IMO we either need to replace the tables with a different layout element or make these two sections full-width.

### DIFF
--- a/content/releases/3-6/fiber/release-guide.txt
+++ b/content/releases/3-6/fiber/release-guide.txt
@@ -55,19 +55,19 @@ When you now click on a link in the Panel, the event is intercepted and a JS fet
 ```
 Fiber now does two things:
 
-1. It updates the URL in the browser
-2. It refreshes the Vue app with the new state from the new Fiber object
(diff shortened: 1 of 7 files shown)


bastianallgeier:
I don't really mind that the table scrolls sideways tbh.

lukasbestle:
It's fine regarding layout and design, but for actually reading it I think it's not optimal.

bastianallgeier:
@lukasbestle totally! It's a trade-off. What about moving version number and throws columns to the right? They are not that important at all.

lukasbestle:
For these tables, yes. I think "Use instead" should be the second column.

There's also "Namespace changes", but it's just about fine still. At least on larger displays.

bastianallgeier (approved):

PR #1565, opened 2021-11-19
Update reference-panelsection.txt

plagasul:
There was a repeated part about 'image: icon', the first was longer but erroneous, I fixed it and deleted the second one.


texnixe:
Thanks!

PR #1569, opened 2021-11-20
Add programmable blueprints recipe

texnixe:


lukasbestle (commented):
Great work! 👍
I've fixed a few typos and code formatting issues, hope it's OK that I committed them directly.

One detail I noticed: Dynamic queries like `page.children.filterBy('tags', '". $tag . "', ',')` and `site.find("' . $page->slug() . '").children` don't escape the dynamic value, so "query injection" attacks are possible. If we assume we cannot fully trust user input, we should escape quotes in the dynamic values. But the question is whether the code needs to be production-ready or if that would blow up the examples too much.

texnixe:
@lukasbestle Thanks for reviewing. What would you do? Feel free to escape or add a note to users as you see fit?

lukasbestle:
@texnixe Sorry for the late reply. To be honest I'm not sure. It somehow smells like another missing core feature. Ideally the query language would support query placeholders that work a bit like SQL prepared statements. Escaping can never be fully reliable and not protecting against attacks at all is also a bad idea. But I'm not sure if such prepared statements would be viable.

As for the recipe: Maybe it's enough to add a warning box at the top that dynamic blueprints can be risky if they handle unvalidated or untrusted user data and that this recipe assumes that all handled content is trusted. We can then still update the recipe later in case we have a core feature for dynamic queries.

lukasbestle:
Also pinging @getkirby/kirby-staff for opinions.

lukasbestle (approved):
I think it's great. Only needs rebasing against `main` and thoughts from other team members on the dynamic queries.

PR #1575, opened 2021-11-23
fix(docs): update k-dialog example to match props

gearsdigital:
I found a very small issue in the `<k-dialog>` example. The attribute to change the submit button text isn't `button` but `buttonText` according to the described properties.


distantnative:
Thanks!

PR #1595, opened 2021-12-03
Add limitations and workarounds with TailwindCSS

anselmh:
I edited the page because I was missing information and [it seems I’m not the first](https://forum.getkirby.com/t/css-classes-in-writer-output/20745). Feel free to edit, adjust or merge.

Generated output can’t easily be used with TailwindCSS so this documents the limitations and provides some ideas for available workarounds.


distantnative:
@chrfickinger same here since this was your original recipe

anselmh:
Do you have any preference / idea of how to deal with this now? I would be able to extend the documentation further but since this one is stale since months I’m not sure if it’s worth the effort.

For example I could add https://helloanselm.com/writings/quick-tailwindcss-production-setup and https://helloanselm.com/writings/using-tailwind-within-css.

texnixe:
@anselmh Yes,  would be great if you can extend this a bit. Maybe also mention the typography plugin which might be useful in block contexts. I think it's mentioned in one of your articles.

texnixe:
I'll merge the current state, then we can add to it when you get round to it.

texnixe (approved):

texnixe:
Changes added via a new branch

PR #1605, opened 2021-12-11
add cookbook recipe: how to make kirby pwa ready

Manuel-Steinberg:
This looks better.


Manuel-Steinberg:
I think there are better options to make Kirby PWA installable. Also, my cookbook uses 3rd party scripts which I do not like any longer. So I am closing this PR.

PR #1619, opened 2021-12-19
Fix typo

sebastiangreger:
`go('login')` does not work, unless such path is configured as a custom route

### DIFF
--- a/content/docs/3_reference/4_objects/cms/0_user/0_logout/reference-classmethod.txt
+++ b/content/docs/3_reference/4_objects/cms/0_user/0_logout/reference-classmethod.txt
@@ -9,7 +9,7 @@ if($user = $kirby->user()) {
   $user->logout();
 }
 
-go('login');
+go('panel/login');
 
 ?>
 ```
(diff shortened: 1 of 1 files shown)


texnixe:
Hm, for frontend login where this code is used, it is not supposed to redirect to the panel login page but to a custom login route. A user with no panel access cannot log in via the Panel login form.

texnixe:
Maybe it would make more sense to add a sentence to the example that a login page is necessary for the example to work.

sebastiangreger:
I see what you mean …indeed not a "typo", just lack of context. I updated the PR; but actually started to think that this last line of the example could as well be dropped here completely?

texnixe:
Well, a bit more context for an example is never bad. I like your updated PR!

texnixe:
Thanks, @sebastiangreger

sebastiangreger:
Any time :bowing_man:

It's not that I am hunting for "typos". But when something in the docs throws me off (I was 100% sure that `go('login')` must have been a thing before, as my copy-paste led to an error :smile:) a PR is only a few seconds to prepare for future-me visiting this docs page again…

texnixe:
> It's not that I am hunting for "typos".

I'm sure you don't and even if, that would be great because even just fixing typos helps us to make the docs better. And even more so, when things do not work as expected and need clarification. So we are really honestly thankful!

PR #1624, opened 2021-12-29
Update Cookbook

Werbschaft:
In this recipe there is still the old method with `published` which does not work anymore and has been replaced by `listed()`.


afbora:
@Werbschaft No, `published` is still active method. `->published()` method returns listed and unlisted children.

Werbschaft:
@afbora you are right, sorry! I thought somehow that is still the old method.

PR #1632, opened 2022-01-09
Add recipe translate exception messages

texnixe:


distantnative:
@texnixe what's your opinion here: is this a legit use case or are the non-translated messages rather a flaw that we should fi in the core?

texnixe:
When we decide to change it, we can remove this recipe again.

PR #1649, opened 2022-01-14
Add user registration recipe

texnixe:
Doesn't make sense to publish it before 3.6.2 because of issues with non-fiber redirects.


afbora:
Then it would be better for this recipe or the required recipes specify the minimum Kirby version requirement.

distantnative (approved):

PR #1655, opened 2022-01-26
Remove deprecated plugins

texnixe:


bastianallgeier:
The auto compression is really getting on my nerves. @lukasbestle @distantnative what's the best way to undo it here? It's not even related to the PR.

distantnative:
@bastianallgeier I'd suggest to really just remove the workflow for now. It annoys me more than it helps

bastianallgeier:
Could this be a manual workflow? I think that would be ideal.

afbora:
I'm not sure but may be you interest: https://imgbot.net/ (I have not used it and do not know all its features. We can consider it as an alternative solution.)

distantnative:
@bastianallgeier manual can be configured, yes: https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

PR #1658, opened 2022-02-02
Add PHP 8.1 support for requirements

afbora:


texnixe (approved):

texnixe:
Wondering if we need a since tag again, though, because 8.1 wasn't supported from 3.6.0?

afbora:
IMHO, No. `Get up and running` page must be valid for the last updated version. For example, it would be pointless to write all past versions for "Browsers for the Panel".

PR #1678, opened 2022-03-16
Fix: The Toolkit `A` class is uppercase

adamkiss:


texnixe:
Thanks, @adamkiss, that looks much better.

PR #1682, opened 2022-03-21
Support nullable parameters for hooks

### LINKED ISSUE #1591: `page.changeStatus:before` hook params mixing when nullable position
`$position` param should be nullable to avoid getting error. But when position nullable as  `?int $position  = null` , document page mixing like following screenshot:

screen-capture-1272-page changeStatus_before - Kirby CMS-getkirby test

distantnative:
Fixes #1591

@afbora could you please do some testing? (if it works, if it doesn't have a regression) that would be super helpful 🙏


afbora (approved):
I've tested on dozens reference pages and seems works great without regression 👍

PR #1685, opened 2022-03-21
Close menu when search is opened

### LINKED ISSUE #1513: Mobile: search hidden by navigation
If you’re on the mobile version, and open the navigation, tap the search opens the search bar under the navigation  and it’s obscured by the nav, which doesn’t close;

expected:
Close the nav or set a higher z-index for the search modal

distantnative:
Fixes #1513

### DIFF
--- a/assets/js/components/menu.js
+++ b/assets/js/components/menu.js
@@ -1,7 +1,6 @@
 export default class {
-
   constructor() {
-    this.$el = document.querySelector(".menu-toggle")
+    this.$el = document.querySelector(".menu-toggle");
 
     if (this.$el) {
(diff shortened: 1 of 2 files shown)


afbora:
Actually works great for nav and search button. But I guess we need more comprehensive enhancement like closing navigation when clicking outside it 🤔

distantnative:
I tried this but it just opened more problems. I think it is ok that you need to explicitly close the menu (not just clicking outside). But we need to avoid that search and menu overlap.

afbora (approved):
So I can confirm that overlapping issue fixed. Let's fix the closing nav issue another time.

PR #1693, opened 2022-03-27
Add cookbook recipe about how to extend core classes

texnixe:


distantnative:
I'm only reading over this on GitHub, but does it really fully work? We often reference classes ourselves explicitly, e.g. `use Kirby\Cms\App` - that wouldn't be overwritten by the custom class, would it?

texnixe:
Yes, it works. `App::instance()` returns the custom class.

PR #1699, opened 2022-03-30
example shows caching

bnomei:
which is plugins focus instead of fetching

### DIFF
--- a/content/plugins/bnomei/boost/plugin.txt
+++ b/content/plugins/bnomei/boost/plugin.txt
@@ -21,10 +21,18 @@ Description: Boost the speed of Kirby by having content files of pages cached, w
 Example:
 
 ```php
-// kirby core
-$page = page($somePageId);
-// faster
-$page = boost($somePageId);
-// fastest
-$page = boost($somePageDirUri);
+// 1) set a cache driver in site/config/config.php
+return [
+    'bnomei.boost.cache' => [
+        'type' => 'apcu', // php, redis, mysql, sqlite ...
+    ],
+];
+
+// 2) add boostid field in site/blueprints/pages/blogpost.yml
+
+// 3) extend model in site/models/blogpost.php
+class BlogpostPage extends \Kirby\Cms\Page
+{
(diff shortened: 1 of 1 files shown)


distantnative:
To be honest with this new example I would have zero clue what this does

bastianallgeier:
Same here to be honest. I think I would keep the preview like it was and wait for when we finally start showing the readme on the plugin page. It will be a far better option to show more details.

bnomei:
but the example you did create shows only the _resolution of relations_. like what autoid does. but thats not what boost is about, its something it can do as well.
i really would appreciate if you let me as the author decide what the preview should contain. if thats not possible then please remove your preview and show none at all.

distantnative:
@bnomei Can you try to explain what you think your proposed example shows? Cause as I wrote, from that I wouldn't understand at all what it does - but I didn't get any reply to that.

bnomei:
it shows part of the setup in adding a trait to a custom page model to extend and overwrite methods. this allows the page to be "boosted" and read/write from a cache instead of just the content file.

the second part shows the config setting where the plugin is told to use the apcu cache driver.

what is important to me is that apart from adding a special field (just like autoid) to the page blueprint these two steps are whats needed to make the page load from cache.

i know its not explaining itself well and i dont care. but these steps are more important to the plugin that the `boost()` helper method the current example shows.

distantnative:
@bnomei i feel like you're approaching this a lot from your perspective as plug-in developer and less from the user's perspective. You have to imagine the user has no clue what your plug-in does. So they aren't yet really interested on what they have to do with your plug-in, but what the plug-in solves for them. And I'd argue that's exactly the boosted lookup that the current example shows.

Adding a trait to a page model doesn't help me at all to understand in the first place what functionality the plug-in provides to me. 

It's like with a microwave: the key part in the first place is not telling the users how to use it (putting plate in, setting timer) but what it does (heating up food).

bnomei:
we agree that we dont agree. please remove the code example.

PR #1700, opened 2022-03-30
example shows magical keys from objects

bnomei:
which is main feature of plugin


lukasbestle:
Hi Bruno, on the detail page it's great, but in the card the text gets cut off because of the long line length.

bastianallgeier:
I think we should probably introduce two examples. One for the card and one for the header of the plugin page.

distantnative:
@bastianallgeier https://github.com/getkirby/getkirby.com/pull/1714 adds a new `preview` field that can be used to define a separate example just for overviews.

PR #1718, opened 2022-04-09
Releases page

distantnative:


texnixe:
Are the half cut-off stars a feature or a bug:

Bildschirmfoto 2022-04-09 um 20 12 09

Looks good for a start, still a few typos. Pity that we don't have release pages for all 3.x versions, looks like something was missing, but it is as it is.

distantnative:
Definitely bug 😂

lukasbestle:
Can we also include all releases individually, including patches? That would fill the gap between 3.0 and 3.5. We would have to add the releases manually, but since we have to update the `kirby` dir on every release anyway, it's not a huge deal. To simplify it, we could store the releases without release notes in YAML in `releases.txt`.

We could also store and display the release date with each version.

distantnative:
I think including all releases, also patches, would be a bit of a disservice. The point is to highlight the features etc. - having dozens of patch releases would hide those again. Then we could continue linking to the GitHub releases list.

texnixe:
I agree with @distantnative and would limit this to the main releases, or at least those that come with notable features.

lukasbestle:
The smaller releases could be single-line entries with just the version number and date. I think they can be quite compact. What this would also allow is to highlight security releases in the list. I know you intended the releases page as a marketing thing, but I feel like we can make it the central hub for all releases, maybe even with an RSS feed in the future.

distantnative:
> central hub for all releases, maybe even with an RSS feed in the future

I think for that the Github releases is already great. Not sure if we need to reinvent that - I'd think the marketing aspect is the one thing missing so far.

texnixe:
Closing because done now

PR #1720, opened 2022-04-11
Home: link to latest release

distantnative:


lukasbestle:
I'm all for showing the release directly on the homepage.

Looking at it I find it a bit too subtle even. A while ago we had a draft with a large colorful "3.6" (back then it was 3.5 probably). With the current layout it could be too much and distract from the screenshot, but maybe we can bring back a bold but faded 3.6 somehow.

distantnative:
I totally agree, at the same time we've moved in circles around this idea for months without coming up with a more visual way that works for us. That's why I would vote for merging at least this subtle one first and not wait until we ever find a more visual way.

lukasbestle:
I agree. Still a *lot* better than nothing.

Maybe let's wait for @bastianallgeier's view on this. If you have an idea for a visual way that's not too much work, we can work on this. Otherwise I'm all for merging this immediately and optimizing it later.

texnixe:
Hm, currently, with the sales notice and the top container I think the additional button makes it look a bit overcrowded. Although I'm all in favor of showing the current release on the home page as well.

distantnative:
Sales are 3-4 of 52 weeks per year, I wouldn't focus too much on that.

texnixe:
@distantnative Since we now have the banner at the top, can we close this PR or is it still relevant?

distantnative:
@texnixe Do we have the banner always there? To me it feels like a very temporary thing, at least from the look. Would be great to permanently get from the home page to all the release note pages as they show off our great recent features

texnixe:
@distantnative Maybe I just thought it was sort of permanent because that banner has been sitting there since beginning of October 🙂. But that's why I was asking.

lukasbestle:
We basically always have a banner, but it may not always be the current release. So I agree with Nico that the button still makes sense. Regarding sales: Maybe we can hide the button when there's an active sale. But on the other hand, that would be confusing for those who want to come back to the release page and then suddenly can't find the button anymore.

I think we really need @bastianallgeier's opinion.

texnixe:
I looked at this again and think we should merge it. Definitely better than having no link to the latest release page at all.

distantnative:
👍 we'd just need to replace 3.6 with 3.9 😂

texnixe:
We'd better parse the main version from `$kirby->version()` instead of handling this manually?

distantnative:
@texnixe I like that

PR #1738, opened 2022-05-01
Improve file types documentation

kevinvennitti:
1. Add link to MDN for common MIME types ([here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types))
2. Add link to the list of file types categories ([here](https://getkirby.com/docs/guide/content/files#example-page-with-files__supported-file-types))
_**Maybe we should explicitly list available values instead: `audio` `code` `document` `image` `video`?**_


texnixe:
Thanks for your PR. Yes, I think a list of possible categories right in the page rather than having to go to another location would make sense.

kevinvennitti:
> Thanks for your PR. Yes, I think a list of possible categories right in the page rather than having to go to another location would make sense.

Done ✅

PR #1741, opened 2022-05-06
Add <?php

christophknoth:
Seems like a small change but if a newbie copies without the <?php the site will be broken and they have no idea why. I encountered this problem a couple of times in workshops.


bastianallgeier (approved):
That's a good point! We often omit it for brevity, but it might really not be a good idea in such cases.

PR #1750, opened 2022-05-23
Add Beyond Kirby guide section

texnixe:
Short intro about different ways to use Kirby „headless“


bastianallgeier (approved):
Super nice!!

PR #1753, opened 2022-05-31
Additions to the "Beyond Kirby" guide

lukasbestle:
From this discussion on Discord: https://discord.com/channels/525634039965679616/978572175013470218/979339042812477450

I've also added an info box on "headless use of headless tech". Please let me know what you think, maybe it's too extreme.

### DIFF
--- a/content/docs/1_guide/19_beyond-kirby/guide.txt
+++ b/content/docs/1_guide/19_beyond-kirby/guide.txt
@@ -35,12 +35,12 @@ With Kirby, you have several options to run in "headless" mode.
 The following is an introduction to each of these options and an overview of their uses.
 
 <info>
-Sometimes, "headless" technologies are used for their own sake, for example to satisfy a buzzword-driven project requirement or to reduce the (perceived) development effort. This is often at the cost of performance (due to additional requests, JS code and frontend calculations), accessibility (due to issues with semantics), efficiency (due to a higher energy usage) and therefore at the cost of the overall user experience.
+The technologies and approaches we describe here have very valid uses as we showed above.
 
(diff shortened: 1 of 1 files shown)


texnixe:
Great additions, what I'm not so happy with are the modified heading levels. A TOC where the possible approaches are visible seems more useful to me.

texnixe:
> The technologies and approaches we describe here have very valid uses as we showed above. Please use them wisely and not ... *headless*. 🙂

This could come across as a bit paternalistic?

lukasbestle:
> A TOC where the possible approaches are visible seems more useful to me.

Good point. The reason I changed it was to add the "Approaches for 'headless' Kirby" heading (to make the first intro section shorter and easier to read). I thought that it makes sense from the semantics that the different approaches are now subheadings. But I agree with you that the missing TOC is not great.

I can reverse the heading level changes so that the options are on the same level as the "Approaches" heading.

> This could come across as a bit paternalistic?

That's what I thought as well. I still think we should bring our opinion on this topic across. How can we phrase it better?

texnixe:
Maybe just turn the argument around, then we don't need the advice at the end?

The technologies and approaches we describe here have very valid uses as we showed above. 

Sometimes, however, "headless" technologies are used for their own sake, for example to satisfy a buzzword-driven project requirement or to reduce the (perceived) development effort. This is often at the cost of performance (due to additional requests, JS code and frontend calculations), accessibility (due to issues with semantics), efficiency (due to a higher energy usage) and therefore at the cost of the overall user experience.

lukasbestle:
I like it. 👍

PR #1759, opened 2022-06-10
Fix two potential problems in the JS code sample

oelna:
Prevent URLs with hashes from messing up the JSON loading with fetch. Omit .innerHTML when inserting the HTML result, so event listeners and other attached code are kept intact.


texnixe:
Thank you!

PR #1773, opened 2022-06-28
staticache (fork) plugin

bnomei:


lukasbestle:
Hi Bruno,

Thanks for creating the fork, it will certainly be useful in the time until Staticache is in the core.

This is our longterm plan for Staticache, which is why we also didn‘t add our own plugin to the list (currently in beta). That‘s why we decided not to add the fork either. Thanks for your understanding.

PR #1779, opened 2022-06-29
Hint about adding custom Panel CSS from a plugin

sebastiangreger:
via https://forum.getkirby.com/t/is-it-possible-to-autoload-a-panel-css-from-a-plugin/15909 – this seems not to be documented so far?


texnixe:
Thanks, @sebastiangreger!

PR #1798, opened 2022-07-17
Update security page and security guide

lukasbestle:
In early preparation for the update check feature.

### DIFF
--- a/content/docs/1_guide/19_security/guide.txt
+++ b/content/docs/1_guide/19_security/guide.txt
@@ -13,7 +13,7 @@ Text:
 
 We constantly improve the code and security of Kirby with every update. You can find the latest Kirby version and security incidents of previous Kirby releases on our (link: security#supported-versions text: security page).
 
-Please always make sure to **update to the latest version** in order to fix possible issues and vulnerabilities.
+Please always make sure to **keep your Kirby version up to date** in order to fix possible issues and vulnerabilities.
 
 <warning>
 However, *never* update a live website directly on the server. Test updates locally or on your staging server first to make sure none of your code breaks or plugins suddenly stop working. Always make sure to have a backup in place.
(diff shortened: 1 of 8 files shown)


content/docs/1_guide/19_security/guide.txt
```diff
@@ -31,6 +31,16 @@ Kirby's current version number is stored in `/kirby/composer.json` in the versio
 
 <info>Make sure to never show the version number publicly in one of your templates as that may give attackers valuable information about your site.</info>
 
+### Use versions with security support
+
+Whenever we find or get notified of a vulnerability, we release updates for all Kirby versions with active support and security support. Active support means that the version receives functionality updates, enhancements and bug fixes. Security support means that the version only receives security updates.
+
+Versions that are "end of life" receive no updates whatsoever. We cannot ensure that these versions are free of vulnerabilities and security researchers generally don't review these versions for vulnerabilities either.
```
texnixe, line 38: What I'm wondering Shouldn't 3.* versions before a security fix then have an EOL tag as well, instead of 3.* => active-support, at least that might be a bit misleading.

lukasbestle: I have written this guide section so that it can be linked to from the security issue list in the Panel System view. So users who get sent here are on an older major release that will not receive any more updates. Whereas users who are on 3.x but on a release with a known vulnerability won't get the EOL warning but a specific warning about the vulnerability instead, also with the recommendation to update.

(Of course the update check will only be implemented in 3.x, so it won't be displayed on v1 or v2. But if we ever release v4 and v3 is EOL, the EOL check will become active. All these thoughts are planning ahead.)

The data in the table is structured for the update check logic. I have updated the guide wording though to be more clear. Is it better now?

texnixe: Yes, it's fine!

distantnative:
@lukasbestle let me know when you're ready and need a review

lukasbestle:
@distantnative It's ready. :)

lukasbestle:
@distantnative Now it's ready. Probably easiest to review each commit separately for cleaner diffs. Maybe review the [docs commit](https://github.com/getkirby/getkirby.com/pull/1798/commits/de123ae89ad497e8c13c0d46f95877baa7a849ae) first as it makes some of the other changes easier to understand.

distantnative (approved):

PR #1800, opened 2022-07-18
Remove cached array, fixes #1799

### LINKED ISSUE #1799: Entries missing (and misnamed) in Validators sidebar navigation
**Issue with Reference**  

Several validators listed on the main [Validators](https://getkirby.com/docs/reference/system/validators/) page are missing from the sidebar navigation when you click through to an entry. For example: 

https://getkirby.com/docs/reference/system/validators/contains

Validators listed on the main Validators page, but missing from the sidebar navigation are:

`accepted`
`contains`
`denied`
`empty`
`endsWith`
`json`
`less`
`more`
`notContains`
`notEmpty`
`required`
`startsWith`
`time`

Also, several validators display the wrong name in the sidebar navigation. I suspect that this has something to do with the use of the page slug, rather than a proper title field.

`maxLength` is listed as `max-length`

https://getkirby.com/docs/reference/system/validators/max-length

`maxWords` is listed as `max-words`

https://getkirby.com/docs/reference/system/validators/max-words

`minLength` is listed as `min-length`

https://getkirby.com/docs/reference/system/validators/min-length

`minWords` listed as `min-words`

https://getkirby.com/docs/reference/system/validators/min-words

`notIn` is listed as `not-in`

https://getkirby.com/docs/reference/system/validators/not-in

**Expected**  

All current validators should appear in the sidebar and display the correct naming conventions.

**Screenshots**  

Case in point, the `contains` validator page is not listed in the sidebar.

Screen Shot 2022-07-18 at 10 48 51 AM

texnixe:


lukasbestle (commented):
I think the cache itself is fine (we also use it in the `reference-class` model for example), but it's never set from the dynamic children list.

Have you tried making the `return` statement at the bottom:

```php
return $this->children = Pages::factory($children, $this);
```

texnixe:
You are absolutely right!

PR #1801, opened 2022-07-25
Update versioning docs

lukasbestle:


content/releases/3-6/changes/panel/link.txt
```diff
@@ -135,6 +135,22 @@ Fixes:
 
 Breaking:
 
+<info>
```
texnixe: Maybe we should move this repeating part to a snippet or a tag for easy integration into new release pages?

texnixe (approved):
Looks good to me, see my suggestion.

PR #1802, opened 2022-07-28
Fix for missing $page variable in complex grouping example

joh-sch:
Change `$page` to `$p` in complex grouping scenario example to fix undefined variable error. The callback of the `map` function in the example is being passed the page object as `$p` (see line 99) but then refers to it as `$page` later in the callback (see line 100 and 102).


distantnative:
Thanks!

PR #1803, opened 2022-07-28
Two minor tweaks and an explanation regarding how to style the content

morrits:
This is about the following Cookbook article: https://getkirby.com/docs/cookbook/setup/kirby-meets-tailwindcss

There were two things in the explanation that were outdated. I have updated this to the current way of use. In addition, I have added information on how to style your content by using the Tailwind CSS Typography plugin.

### DIFF
--- a/content/docs/2_cookbook/9_setup/0_kirby-meets-tailwindcss/cookbook-recipe.txt
+++ b/content/docs/2_cookbook/9_setup/0_kirby-meets-tailwindcss/cookbook-recipe.txt
@@ -8,7 +8,7 @@ Published: 2021-08-09
 
 Text:
 
-This is a walkthrough on how to install and use Tailwind CSS with Kirby.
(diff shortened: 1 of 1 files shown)


morrits:
Small addition:

In the text below, '--purge' should also become '--content'. I don't know exactly how to modify my commit. Excuse!

```
### Explanation
Use `watch` to observe changes and generate a CSS file on change
Use `build` to generate a final minified CSS file
- `-i ./src/css/tailwind.css`: input file with Tailwind's css classes
- `-o ../assets/css/styles.css`: output file which will be generated
- `--purge '../site/**/*.php'`: folder to watch for Tailwind's classes
- `-w`: defines watch mode
- `-m`: minify output file
```

texnixe (approved):
Since this can happen with every update, we should probably add some information about the Tailwind version this refers to.

PR #1808, opened 2022-08-10
Correct link to Extending core Blocks

ovenum:
The link to Extending core blocks cook book recipe is returning a 404. 
Now it links to the correct location.


afbora:
Thanks

PR #1809, opened 2022-08-11
root URL warning in the autodetect feature

adamkiss:
If you use "root urls for the default language" and then, later, try to setup automatic language detection, you might overlook the fact that it won't work, because you're **focusing on the detection thing**, not the fact that you **disabled it by the other setting**.

This might save someone _hours_ in the future. Not that anybody would waste hours on overlooking such thing, _of course_.


adamkiss:
👏

PR #1811, opened 2022-08-14
chore: add updated description and image to the enhanced-link-dialog-plugin

gearsdigital:
After updating my Plugin [Enhanced Toolbar Link Dialog](https://github.com/gearsdigital/enhanced-toolbar-link-dialog) I changed the Plugin UI and description slighly. This MR will update the plugin site with the values.


content/plugins/gearsdigital/enhanced-toolbar-link-dialog/plugin.txt
```diff
@@ -14,4 +14,4 @@ Subcategory: enhancement
 
 ----
 
-Description: The Enhanced Toolbar Link Dialog makes it easy to add internal or external references to your content by extending the default link dialog of Kirby's textarea fields
+Description: Extend Kirby's default link dialog with a search functionality to create links to existing or external pages right at your fingertips.
```
distantnative, line 17: I like the rephrasing except that now it doesn't mention anymore that this is in context of the textarea field. Would be good to make that relation.

texnixe: @distantnative The point of the new version is that it is no longer limited to the textarea field.

distantnative: Ok, got it :D

gearsdigital: @texnixe, correct, that was my intention. Thanks for accepting this MR!

PR #1820, opened 2022-09-07
Correct info on slug field for default language

trych:


content/docs/1_guide/8_languages/3_translating-urls/guide.txt
```diff
@@ -54,8 +54,8 @@ and the URL for the German translation will be:
 http://yourdomain.com/de/projekte
 ```
 
-You don't have to add the `slug` field to your text files for your default language. Kirby will use the folder name in that case.
+The `slug` field in your text files will not have any effect for your default language. Kirby will always use the folder name for the default language.
```
distantnative, line 57: To be honest, I would only change the second sentence as you suggest. The first one becomes more confusing with your update as many users won't have any slug field in the first place (and then ask themselves, what slug field?)

texnixe: But we don't even support the slug field for the default language anymore like we used to? So the first sentence in its current state is more confusing, I think.

trych: Agreed. Will prepare another PR.

PR #1822, opened 2022-09-09
Add `sort` option to page blueprint doc

jhogervorst:
I couldn’t find a way to disable certain users from sorting a page. Turns out that I can set the `sort` option on the page blueprint, but it was missing in the docs.


texnixe:
Thanks!

PR #1832, opened 2022-09-18
Extend plugin JSON template for update check

lukasbestle:
This enables the Panel update check for plugins.

I suggest to start simple with hardcoded update data. Once we have the plugin hub, we can make all of this configurable so that plugin devs can e.g. also submit vulnerabilities and control which versions are free or paid upgrades.

### DIFF
--- a/site/models/plugin.php
+++ b/site/models/plugin.php
@@ -25,26 +25,49 @@ class PluginPage extends Page
         return $this->images()->findBy('name', 'card');
     }
 
-    public function download()
+    public function changes($version = null)
+    {
+        if ($this->content()->has('changes')) {
+            return $this->content()->get('changes')->value();
+        }
+
+        $url = $this->repository()->value();
+
+        if (Str::contains($url, 'github')) {
+            if ($version) {
+                return $url . '/releases/tag/' . $version;
+            }
+
+            return $url . '/releases/latest';
+        }
+
+        return $url;
+    }
+
+    public function download($version = null)
     {
         if ($this->content()->has('download')) {
             return $this->content()->get('download')->value();
         }
 
-        $url   = $this->repository()->value();
-        $branch = $this->branch()->value() ?? 'master';
+        $url    = $this->repository()->value();
+        $object = $version ?? $this->branch()->value() ?? 'master';
 
         if (Str::contains($url, 'github')) {
+            if ($version) {
+                return $url . '/archive/refs/tags/' . $version . '.zip';
+            }
+
             return rtrim(Str::replace($url, 'github.com', 'api.github.com/repos'), '/') . '/zipball';
         }
 
         if (Str::contains($url, 'bitbucket')) {
-            return $url . '/get/' . $branch . '.zip';
+            return $url . '/get/' . $object . '.zip';
         }
 
         if (Str::contains($url, 'gitlab')) {
             $repo = basename($url);
-            return $url . '/-/archive/' . $branch . '/' . $repo . '-' . $branch . '.zip';
+            return $url . '/-/archive/' . $object . '/' . $repo . '-' . $object . '.zip';
         }
 
         return $url;
@@ -136,8 +159,8 @@ class PluginPage extends Page
 
     public function toJson($onlyIfCached = false)
     {
-
         $screenshot = $this->images()->findBy('name', 'screenshot');
+        $version    = $this->version($onlyIfCached);
 
         return [
             'title'  => $this->title()->value(),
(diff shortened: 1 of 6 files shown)


site/models/plugin.php
```diff
@@ -151,7 +174,29 @@ public function toJson($onlyIfCached = false)
...
+                '*' => [
+                    'description' => 'Actively supported',
+                    'latest'      => $version,
+                    'status'      => 'active-support'
+                ]
+            ],
+            'urls' => [
+                '*' => [
+                    'changes'  => $this->changes('$version'),
+                    'download' => $this->download('$version'),
```
distantnative: I think these two are not supposed to be wrapped as strings?

lukasbestle: They are. `$version` is a placeholder in the URL template.

distantnative: Why are we doing it like this and not passing the actual version here? This. is very confusing to read tbh - especially sine `::changes()` apparently is also supposed to be called with an actual version number?

lukasbestle: The update checking code can use the URL template to build URLs for any version, not just the latest one. This is e.g. necessary in the following cases:

- If there is a paid upgrade but the update check wants to suggest the last free update.
- If there are vulnerabilities and the update check wants to suggest the minimum possible security update.

I've added a comment for future reference.
The format of the security data is also documented here: https://github.com/getkirby/getkirby.com/blob/de11bdecb1542cfccba3314ac994baafba8e9efa/content/security/security.txt#L38

distantnative: I would suggest using `'{{ version }}'` as that's our common string template placeholder, that also works directly with `$helper.string.template()`.

bastianallgeier: I agree about the placeholder for version. The rest looks good to me.

lukasbestle: @distantnative @bastianallgeier I tried it, but frankly it's becoming a mess. We already use the placeholder syntax in our security data so that getkirby.com can fill some information dynamically:

https://github.com/getkirby/getkirby.com/blob/b7b296e11b105190ba71cb9d7a5028597aaf8db7/content/security/security.txt#L179-L191

If we also use the same syntax for the placeholders that should be kept, it's going to be really confusing IMO.

distantnative: Why would it be a mess? For me the referenced lines would be rather an argument to make it consistent. Right now we have to different ways of placeholders in the same place - to me that's the really confusing part. One consistent style (and I think there is a good argument for `{{ }}` und the Kirby world) would be much better.

distantnative: `Str::template()` also supports only partial replacement. So getkirby.com can already replace what it needs to replace and the Panel fills in the remaining placeholder.

bastianallgeier: @lukasbestle I guess your template PR is exactly about solving this problem, right?

lukasbestle: > @lukasbestle I guess your template PR is exactly about solving this problem, right? 

Yes, exactly.

lukasbestle: I just find it confusing that we'll have some placeholders that are replaced by getkirby.com and some placeholders that are left over for dynamic replacement. Down the line we won't know which are which without looking it up.

I'm fine with the change, but then I'd suggest to use a custom format for the "static" placeholders, e.g. `[{ latestMajor }]`. What do you think?

lukasbestle: But now that I think about it: Does the `$field->replace()` method even support custom placeholder start and end strings?

distantnative: Can't we just add a normal code comment, explaining which is replaced where? I think that would be much better than a custom format where it's just a different kind of confusing

lukasbestle: That would be in the content file. Since it's a YAML field, this could work. I'll try it.

distantnative: Or even in the system class where it then gets processed.

lukasbestle: I think there is a misunderstanding. This is all just about getkirby.com. The CMS implementation is no problem at all.

distantnative: That's evident. But I don't think it would be a huge problem if there is not comment in the content file. If there is a comment in the PHP code that replaces the first placeholder, that's fine. If there is just a comment on the side of Kirby that actually interprets the input, that'd be totally fine as well. Those placeholders are quite self-explanatory as placeholders if we stick to our code style. And then it's actually not so hard to follow where what gets replaced.

lukasbestle:
I'm merging the PR so I can create the update check PR. I think besides the query templates (which are now implemented as discussed) everything was already fine.

distantnative:
Would've been great if you had given us a little time for a review: We must not modify Kirby core files besides when updating to an actual release version. The problem now is that margining this PR has generated false docs that claim to be for Kirby 3.7.5 while actually that's not how those methods work in 3.7.5.

distantnative:
Actually I think it broke the whole Reference 😱

PR #1836, opened 2022-09-23
Use SVG favicon for dark color scheme

afbora:
image (11)


afbora:
Or should we keep both PNG and SVG?

PR #1843, opened 2022-10-02
3.8 release page and docs

lukasbestle:
## Docs

- [x] https://github.com/getkirby/kirby/pull/4499
- [x] PHP 8.0 requirement
- [x] https://github.com/getkirby/kirby/pull/4580
- [x] https://github.com/getkirby/kirby/pull/4627
- [x] https://github.com/getkirby/kirby/pull/4676
  - `update` option
  - Privacy guide
- [x] Separate more detailed pages for migration guide
- [x] https://github.com/getkirby/kirby/pull/4612
- [x] Permalinks
- [x] https://github.com/getkirby/kirby/pull/4640
- [x] Translate options via i18n key https://github.com/getkirby/kirby/issues/3955
- [x] New field options syntax @distantnative 
- [x] New command extension api https://github.com/getkirby/kirby/pull/4765


content/docs/1_guide/15_plugins/4_plugin-setup-autoloader/guide.txt
```diff
@@ -30,14 +30,16 @@ The most efficient way to load PHP classes is to use an autoloader. It will dyna
...
 ## How to use an autoloader in your plugins
 
-Kirby ships with the (link: docs/reference/templates/helpers/load text: `load` helper), which is a super simple class autoloader.
+Kirby ships with the `F::loadClasses()` method, which is a super simple class autoloader.
 
 You can call it in your plugin's `index.php` with the list of your classes to load:
 
 ```php "index.php"
-load([
+use Kirby\Filesystem\F;
```
distantnative: I feel like we are raising the bar here for plugin developers that aren't super versed with PHP - now they also need to understand namespaces and the `use` syntax.

lukasbestle: I agree. But now that helper functions can be overridden since 3.7, there's not really another option. Using the helper function in a plugin is not reliable. In the release notes for 3.7 we even added a recommendation for plugin devs to migrate to the class methods.

distantnative: But `F` is globally aliased, so we could then at least use it without the `use` line

lukasbestle: You are right! Didn't get to it today. We also need to update the Pluginkit with the same change.

PR #1844, opened 2022-10-08
Fix link to basic contact form line 7

giannirose:
The "basic contact form" link on line 7 should be: https://getkirby.com/docs/cookbook/forms/basic-contact-form


texnixe:
Thanks for fixing.

PR #1845, opened 2022-10-11
Add fabianmichael/meta plugin

fabianmichael:
I’ve added my metadata plugin to the plugin repo. Please have a look at the `seo` category before merging, as you probably want to update the arrangement of plugins:
 
Bildschirmfoto 2022-10-11 um 12 17 57


afbora (approved):
Plugin registry is compatible for 3.8 👍

PR #1856, opened 2022-10-21
Added fields: to structure reference page example

duness:
Added `fields:` to be congruent with the other examples on that same page. Users who quickly copy and paste to try out this blueprint might forget to manually add `fields:` for this single example.


texnixe:
Makes sense! Thanks!

PR #1862, opened 2022-10-24
image: color and image: back don't  work when adding a semicolon Semi…

vincentrouleau:
…colon

`image: color:` and `image: back:` don't work when adding a semicolon Semicolon.


texnixe:
Thank you!

PR #1864, opened 2022-10-28
update ubuntu version to add PHP8

rlb222:
~/docker-example-2/Dockerfile 
Changed version of ubuntu to ubuntu:22.10 in order to install php version 8+  this php version is needed otherwise Kirby 3.8 will not run. This script as it was will give an error on the 'localhost' kirby page. 

(I am a novice at this, maybe there is a better way to make sure php8 is used. This worked for me)


texnixe:
Thanks for finding this. There were actually two instances where this had to be changed, and I changed it to `latest` now just in case.

rlb222:
Thank you!(& That was quick !!)René —//—Op 28 okt. 2022 om 18:47 heeft Sonja Broda ***@***.***> het volgende geschreven:﻿
Thanks for finding this. There were actually two instances where this had to be changed, and I changed it to latest now just in case.

PR #1868, opened 2022-10-31
Fix a tag bracket

linusrogge:


linusrogge:
sorry, didn't check till the end of the line :D

PR #1872, opened 2022-11-04
Replace hyphens by camelCase in example of line field

jhogervorst:
According to [1], field names should not use hyphens:

> You can only use **alphanumeric characters** and **underscores** in field names.

In my opinion it makes sense to follow that rule in this example. (Even though a line might not be a 'real' field with data, you could be putting users on the wrong foot with such an example.)

[1] https://getkirby.com/docs/guide/blueprints/fields#naming-fields

### DIFF
--- a/content/docs/3_reference/3_panel/3_fields/0_line/reference-article.txt
+++ b/content/docs/3_reference/3_panel/3_fields/0_line/reference-article.txt
@@ -33,16 +33,16 @@ fields:
 
   …
 
-  line-a:
+  lineA:
     type: line
 
   …
 
-  line-b:
+  lineB:
     type: line
 
   …
 
-  line-c:
+  lineC:
     type: line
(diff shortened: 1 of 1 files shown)


texnixe:
Then better replace with camelCase(lineA) for consistency as we don't use underscores in fieldnames anywhere.

jhogervorst:
@texnixe Done!

Is there a reason for preferring camelCase? If so, perhaps that could be added to the docs (section about [naming fields](https://getkirby.com/docs/guide/blueprints/fields#naming-fields)). I was already wondering about which option to pick, and if there is a good reason to prefer one over the other, that would be nice to know as user.

texnixe:
That is totally up to you in your own code, therefore I wouldn't put this in the docs. But we are nowhere using underscores in field names in the documentation.

PR #1873, opened 2022-11-07
Note: animals.csv should use a semicolon separator

giannirose:
I am in the U.S. and the default separator for .csv files is a comma. I used your animals.csv example to render my own spreadsheet. The virtual pages were not created. However, they were listed when I changed my .csv separator to a semicolon as in your example (using find and replace). For some reason my .csv files also had quotation marks around each full entry and others around items with a $ sign. To get the example working, I did not use my entries with $ signs and eliminated all quotation marks.

### DIFF
--- a/content/docs/1_guide/17_virtual-pages/4_content-from-csv/guide.txt
+++ b/content/docs/1_guide/17_virtual-pages/4_content-from-csv/guide.txt
@@ -16,6 +16,10 @@ Text:
 
 In this example we are going to read a list of (dummy) animals from a (file: animals.csv text: CSV file) and create virtual pages for each animal.
 
+<info>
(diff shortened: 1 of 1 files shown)


texnixe:
I changed your PR wording, it is not necessary to use a semi-colon separated file.

giannirose:
A few hours after I wrote I realized there had to be a better way, just as you suggested. Sorry! I'm not experienced to know that but I'm glad I figured it out— I used a components.csv file.

` $csv      = csv($this->root() . '/components.csv', ','); `  

My first method of finding and replacing my commas with semicolons had the unfortunate effect of changing dollar amounts from, for example, "$1,800" to "$1;800" and we wouldn't want that.

PR #1874, opened 2022-11-08
Update My first Panel area docs

bensanlau:
While doing the tutorial, I found the review list doesn't render unless items are passed into `<k-items>`.


texnixe:
Thanks, that might have changed since the recipe was written, can't test this now, but will merge.

PR #1875, opened 2022-11-09
Add separate file example

kevinvennitti:
I put these two code blocks in the `<info>` block: maybe? maybe not?


texnixe:
I don't think that's a good idea. I think it would make more sense to expand the docs under https://getkirby.com/docs/reference/plugins/extensions/page-models and then link to that page.

texnixe:
https://github.com/getkirby/getkirby.com/pull/1879

PR #1877, opened 2022-11-10
Add new plugin "anselmh/kirby-uniform-turnstile"

anselmh:
Let me know if you need anything else.


content/plugins/anselmh/kirby-uniform-turnstile/plugin.txt
```diff
@@ -0,0 +1,32 @@
...
+```html+php
+<form action="<?= $page->url() ?>" method="post">
+    <!-- ... -->
+
+    <?= csrf_field() ?>
+    <?= turnstileField() ?>
+    <input type="submit" value="Submit">
+</form>
+<?= turnstileScript() ?>
+```
```
afbora: This file should be in `uniform-turnstile` directory instead of `kirby-uniform-turnstile` to stable work for plugin update check. Because your plugin registry is `anselmh/uniform-turnstile`: https://github.com/anselmh/kirby-uniform-turnstile/blob/main/index.php#L7

anselmh: Thanks, you’re right! Fixed it now.

afbora:
Thanks for the plugin 👍

PR #1878, opened 2022-11-11
Add warning for deprecated function strftime

olach:
Additional instructions needs to be written. Maybe a strftime replacement will be integrated in a future Kirby version?


texnixe:
We already have a replacement with the intl date handler, see date config options. I suggest we should remove `strftime` completely here instead of putting a warning and use the new date handler as example instead.

lukasbestle:
I agree with Sonja. Replacing the existing example with `intl` makes a lot of sense going forward.

texnixe:
Closing this in favor of https://github.com/getkirby/getkirby.com/pull/1880

PR #1886, opened 2022-11-15
Update virtual pages guide, fixes #1884

### LINKED ISSUE #1884: Update Virtual pages guides with UUID warning
**Issue with Reference**  
Any virtual pages guide/docbook is currently broken, in the sense that if a programmer doesn't add a **self generated uuid** to the content array of the virtual page, kirby generates the uuid and that triggers the creation of an actual page - folder and file in the filesystem.

**Expected**  
"Since 3.8.1: You must include a 'uuid' in your content file"…

**Additional note**  
I'm emphasising "self generate" uuid, because the first thing I tried was of course `'uuid'=>Uuid::generate()`, but that seems to trigger the file/folder creation as well.

texnixe:


lukasbestle:
TBH I'm not sure if the current core behavior is actually a bug. Suddenly saving a virtual page to disk just because it's missing a UUID feels weird. Maybe we should instead throw an exception if a dev tries to use any UUID-related feature with a virtual page that doesn't have an explicit UUID set.

Your changes to the guide solve the issue of saving virtual pages to disk, but they break UUIDs, so it feels like a hack to me.

@distantnative?

distantnative:
I can relate to @lukasbestle comment. But I don't have a good idea how we would identify a virtual page exactly. Would this be something like "throw an exception if no uuid is set directly AND also there is no actual content file"?

texnixe:
If I don't set a UUID for my virtual pages, what happens if I try to reference them in a pages field? Also wondering if we should at least put a warning of some sort into the guide, given that the current guides are all broken when UUIDs are not disabled.

lukasbestle:
Currently those fields will probably break. We could solve this with an automatic fallback to the old-style ID. But I wonder if it makes sense to go that route. Virtual pages can only integrate into the system properly if they contain UUIDs.

So my current opinion is: Devs should be responsible to give UUIDs to their virtual pages – ideally persistent ones.

If you agree, then let's merge this PR. I think it explains it very well with the warning boxes.

distantnative (approved):

PR #1895, opened 2022-11-27
fix error if card page level is > 1

hariom147:
`.replaceAll('/', '+')`only replaces the first slash with a plus in the api call, leading to an error if the selected page is nested deeper than level 1. Using `.replaceAll()`fixes that.


texnixe:
Thank you!

PR #1896, opened 2022-11-28
Fix incorrect instructions

MoritzLost:
The documentation currently says the constant needs to be set to `true` to disable the helper method, but it actually needs to be set to `false`.


texnixe:
Thanks!

PR #1901, opened 2022-12-07
Improve documentation for the invalid() function

MoritzLost:
- Document the syntax variants for the `$rules` array that is passed to the `invalid()` function.
- Add an example for the `in` validator using a nested array
- Improve existing examples (`minLength`/`maxLength` instead of `min`/`max` for strings).


texnixe:
Array syntax should all be converted to short form, the long form partly used is a leftover from old documentation.

MoritzLost:
@texnixe And here I tried to match the existing style 😅 I've updated the document to use the modern array syntax consistently!

texnixe:
Thanks, @MoritzLost

PR #1913, opened 2022-12-19
Chunky plugin for getkirby.com

Foddy:


afbora:
You need to do:

- Add `plugin-developer.txt` file in `/content/plugins/dgtlone` directory.
- `chunky` (`/content/plugins/dgtlone/chunky`) directory name should be `kirby-chunky` (``/content/plugins/dgtlone/kirby-chunky``) because of your registry (https://github.com/dgtl-one/kirby-chunky/blob/main/index.php#L15)

Foddy:
Makes sense, sorry... I just requested a new merge request.

PR #1924, opened 2023-01-04
Update plugin descriptions for author `johannschopplich`

johannschopplich:
One thumbnail and some texts were outdated, I updated them.


content/plugins/johannschopplich/hashed-assets/plugin.txt
```diff
@@ -14,12 +14,12 @@ Subcategory: cachebusting
...
-Description: File name hashes support for the css() and js() helpers
+Description: File name hashes support for css() and js() helpers. Without rewrite rules!
 
 ----
 
 Example:
 
 ```php
-<script src="/js/main.9ad649fd.js">
+<?= js('assets/js/main.js') ?>
```
lukasbestle, line 24: I think this no longer shows what the plugin will do. Maybe we could do something like this:

```php
<?= js('assets/js/main.js') ?>

<!-- will become -->
<script src="/js/main.9ad649fd.js">
```

johannschopplich: Great catch! I've updated it to:


```php
<?= js('assets/js/main.js') ?>

<!-- will become -->
<script src="https://example.com/assets/js/main.9ad649fd.js"></script>
```

… Like it's transformed by the plugin.

PR #1927, opened 2023-01-08
kirby-stripe plugin

tristantbg:


afbora:
@tristantbg Your plugin registry is inconsistent like `tristantbg/kirby-stripe`, `tristanb/kirby-shopify` and your getkirby vendor name is `tristanb`. I think getting it all in order would be great for users to keep up with updates.

afbora:
@tristantbg I think just updating `kirby-stripe` plugin vendor name to `tristanb` seems enough

tristantbg:
@afbora  I never created this tristanb folder and there was also a problem with kirby-shopify plugin at that time.
Maybe we should rename the kirby folder to tristantbg ?

afbora:
@tristantbg I can update your vendor name in getkirby but your `shopify` plugin registry name is `tristanb/kirby-shopify` and you need to update with `tristantbg/kirby-shopify`.

afbora:
@tristantbg I've updated your vendor name. Now you should always use `tristantbg/*` in the registry name of plugins.

tristantbg:
> @tristantbg I've updated your vendor name. Now you should always use `tristantbg/*` in the registry name of plugins.

Ok thanks !

PR #1939, opened 2023-02-08
Show README.md from GitHub

plain-solutions-gmbh:
Sometimes it's useful to see the documentation direct from the GitHub-Repository on a plugin page. That's why i create this feature.

**Before you merge it, let me mention something:**
On each call of a plugin page, getkirby.com makes **up to 12 requests** on the GitHub server. I thing this could be problematic. GitHub could see this as a crawling attack and put getkirby.com on a black list. There's tree ways to avoid this:

1. Put the request script on the page. So it's gonna be cached on the first page call. Problem: The content gets obsolete.
2. Create a script, that crawls the readme files at regular intervals.
3. Using repository hooks (Get a push on repo changes). But this needs some interventions from all plugin users.

The core team should decide, if it's smart to merge this branch.

### DIFF
--- a/site/controllers/plugin.php
+++ b/site/controllers/plugin.php
@@ -8,16 +8,75 @@ return function ($page) {
     ->filterBy('category', $page->category()->value())
     ->not($page);
 
+  $getFromGitHub = function(Array $paths) {
+
+    $request = Remote::request(implode('/',$paths));
+    
+    if ($request->code() == 200) {
+      return [
+        'url' => $request->url(),
+        'content' => $request->content(),
+        'root' => implode('/',array_slice($paths, 0, -1))
+      ];
+    };
+
+  };
+
+  $readme = function () use ($getFromGitHub){
+
+    if (page()->repository()->isNotEmpty()) {
+
+      $repo_path = 'https://raw.githubusercontent.com/'.Url::path(page()->repository());
+      $result = array();
+      $variations = [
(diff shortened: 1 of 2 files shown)


distantnative:
I Like the initiative. We've tried this ourselves before but didn't yet manage to do this properly due to a few hurdles.

Besides the caching, relative paths can be tricky. But I see you tried to solve this. But what about malicious content that could be included in a README? we'd need to protect ourselves from such code just being output and/or executed on getkirby.com

plain-solutions-gmbh:
Maybe an iframe could solve it..?

plain-solutions-gmbh:
Hm. Nope, Iframe isn't a good option:

1. CORS Problems
2. iframes are pretty obsolete. (Ah, It reminds me of my first website of the 90s. A Nav and a iframe. 😇)
3. Markdown isn't readable by browsers.

Even if we escape scary htmls. Still got some broken anchor-links.

Nice idea but not realizable. Should we burry it?

lukasbestle:
> Nice idea but not realizable. Should we burry it?

To be honest I agree. At least for now. We still have plans to allow managing plugin information via https://hub.getkirby.com. This will still take a while as there are other more urgent projects, but with that we could allow plugin devs to extend their plugin pages much more.

In any case, thank you very much for your PR. It was worth a try because having more information on the plugin pages is definitely something we are interested in.

PR #1942, opened 2023-02-14
Cookbook kiosk setup

HashandSalt:
Fix stray chrachter causing markdown formatting in the wrong place.


texnixe:
Fixed directly on the main branch

PR #1950, opened 2023-02-23
replaced wrong heading and added one

opafritz:
replaced Bad with a wrong Good title and added an additional Good one


texnixe:
I don't agree, the two examples are not equivalent.

PR #1961, opened 2023-03-01
add moritzebeling/kirby-favicon plugin

moritzebeling:
Add Kirby Favicon Plugin
https://github.com/moritzebeling/kirby-favicon

### DIFF
--- a/content/plugins/moritzebeling/kirby-favicon/plugin.txt
+++ b/content/plugins/moritzebeling/kirby-favicon/plugin.txt
@@ -1,0 +1,25 @@
+Title: Kirby Favicon Plugin
+
+----
+
+Repository: https://github.com/moritzebeling/kirby-favicon
+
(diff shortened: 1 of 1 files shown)


moritzebeling:
I opened a similar PR last week but just realize that it also had some other stuff in the history (other previously added plugins), so I am recreating this PR with a brand new and clean fork of getkirby.com

content/plugins/moritzebeling/kirby-favicon/plugin.txt
```diff
@@ -0,0 +1,25 @@
+Title: Kirby Favicon Plugin
```
afbora: I think just `Favicon` would be fine.

moritzebeling: Okay :)

PR #1963, opened 2023-03-03
Update Batch updating content cookbook-recipe.txt

mrflix:
The blocks object has to be converted to an array when storing it via `$page->update()`.

### DIFF
--- a/content/docs/2_cookbook/2_content/0_batch-update/cookbook-recipe.txt
+++ b/content/docs/2_cookbook/2_content/0_batch-update/cookbook-recipe.txt
@@ -192,7 +192,7 @@ foreach($collection as $page) {
 
     // update the content file
     $page->update([
-        'text' => $blocks
+        'text' => $blocks->toArray()
     ]);
     
 }


afbora (commented):
All of the following is a valid use and works:

````php
'text' => $blocks

'text' => $blocks->toArray()

'text' => json_encode($blocks->toArray())
````

**Comment by mrflix** [comment]
Hm... I was trying the first option and it didn't work.
Eventually I found [this hint](https://forum.getkirby.com/t/create-custom-blocks-programmatically/22137/12) by @texnixe that the `$blocks` has to be converted to array.

**Comment by mrflix** [comment]
By the way thank's for checking it out so fast, @afbora ☺️

**Comment by afbora** [comment]
@mrflix I have now tried all 3 options and they all work 🤷‍♂️

**Comment by mrflix** [comment]
Maybe it depends on how the blocks got created?

In my case I iterate through already created blocks and add data to individual blocks by converting them to array and then re-creating the block:

```yml
title: Form
icon: edit
fields:
  fields:
    label: Fields
    type: blocks
    pretty: true
    empty: No fields
    fieldsets:
      input:
        name: Input field
        icon: edit
        preview: fields
        wysiwyg: true
        label: false
        fields:
          title:
            type: text
            width: 1/3
          description:
            type: text
            width: 2/3
          text:
            type: textarea
            buttons: false
```

```php
$fields = new Kirby\Cms\Blocks();

foreach($page->fields()->toBlocks() as $block){
	if($block->type() == 'input'){
		$old = $block->toArray();
		$old['content']['text'] = get($block->id());
		$fields->add(new Kirby\Cms\Block($old));
	} else {
		$fields->add($block);
	}
}

$kirby->impersonate('kirby');

$page = $page->update([
	'fields' => $fields->toArray()
]);
```

In this case, when I don't use `toArray`, it mushes the blocks together into one block with the type text (instead of type input).

Edit: Kirby v3.9.1

afbora:
The recipe works, but your example does not. I have no idea 🙈

mrflix:
There is a 🐞 hidden somewhere 😬

texnixe:
Can we close this?

afbora:
Recipe works great, but;

- We can close this PR to show different example use case.
- We can merge this PR so that it always works for different uses.

The decision is yours @texnixe

PR #1970, opened 2023-03-20
Move optimise-it article from Notion to CV

tobimori:
I moved the article from Notion to Read CV.


lukasbestle:
Thanks for updating the site. 👍

PR #1973, opened 2023-03-24
Markdownplus

pgfactory:
New plugin called "MarkdownPlus".


lukasbestle (commented):
Thanks for your contribution.

There is currently a mismatch between your plugin path in this pull request (`pgfactory/markdownplus`) and the plugin name in your `index.php` (`usility/markdownplus`). This would mean that Kirby's update check will not be able to check for updates of your plugin.

pgfactory:
Sorry about that, bug is fixed now.
I hope I didn't overlook another one.

PR #1981, opened 2023-03-26
Add CLI plugin, add "new" filter

texnixe:


distantnative:
It seems to work well for me.
Personally, I am not sure how useful the "new" filter is based on a manually entered publish date. Wouldn't people rather go for most popular, recommend or even official (missing that one really) plugins than just for new?

But I trust your judgement there 👍

texnixe:
@distantnative We already have the featured category, but I find it quite useful to be able to see what's new, just to discover something. But a new category for official plugins is a good idea.

distantnative (approved):

PR #1982, opened 2023-03-26
Enhance archive with links to different Kirby doc versions

texnixe:
The link to v35.getkirby.com is broken, though.


distantnative:
👍 I would though move the version data from the controller to a new config file. I think it could be useful for other features to and then not to repeat ourselves but have it in one place centrally.

distantnative (approved):

PR #1985, opened 2023-03-27
Add new official category in top sidebar

texnixe:


distantnative:
👍 but "Team plugins" doesn't sound right to me. Sounds too much like "Plugins for teams". "Official plugins" not good?

distantnative (approved):

texnixe:
I had "official" first but then changed it again, now back to official

PR #1986, opened 2023-03-27
Add page for breaking changes

texnixe:


distantnative:
Nice!

Some quick thoughts:
- Do ew need `changelog` layout for anything else? If not, I would rather keep it all in the template (instead of one-usage layouts)
- Just the main title would be fine and maybe a bit leaner than those two lines
- Should this maybe be rather in `/releases/changelog` than on the top level?

texnixe:
Ok, I changed the title.

Not sure about adding the page as a subpage of releases, if we want to do this, we would have to exclude it from the `releaes` pages.

Regarding the layout, ok, I was a bit lazy and copied it from the cookbook example, but the format might actually be useful for other pages.

distantnative:
But then we might want to find a more generic name for the layout. With the current name, nobody will dare to use/touch it again. And then I think we shouldn't do the single-use layout.

And on the location, it's ok then. I just feel we're crowding our top-level so much. It's getting hard to keep an overview.

texnixe:
@distantnative You are right, I could actually reuse the article layout, still need to find my way around this layout stuff.

distantnative (approved):

PR #1995, opened 2023-04-03
Add lukaskleinschmidt plugins

lukaskleinschmidt:
Let me know if the example for the types plugin is too much

### DIFF
--- a/content/plugins/lukaskleinschmidt/laravel-vite/plugin.txt
+++ b/content/plugins/lukaskleinschmidt/laravel-vite/plugin.txt
@@ -1,0 +1,33 @@
+Title: Laravel Vite
+
+----
+
+Repository: https://github.com/lukaskleinschmidt/kirby-laravel-vite
+
+----
+
+Category: templating
+
+----
+
+Subcategory: assets
+
+----
+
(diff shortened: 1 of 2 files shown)


lukasbestle:
The code examples are used in the plugin's card and room is very limited. You can preview how it will look like by setting up this repo on your local dev server.

lukaskleinschmidt:
I did. And it works well I would say. Was just not sure about adding a custom css animation is something you would want to add

texnixe:
I know it's not documented yet, but could you also please add a `published` field (Published: 2023-04-04), so that the plugin appears in the `new` category? Thanks!

lukasbestle (commented):
Thanks for adding the metadata. 👍

Just looked closer at the code example of the types plugin and tested it locally. Now I know why it fits into the preview card. 🤯
It's a nice idea to use highlighting and the animation. However I feel like the animation would be too distracting if other plugins also used animated cards. Also the text colors don't really fit with the site colors.

So I suggest the following: No animation and colors from the CSS variables. Would that be fine for you?

lukaskleinschmidt:
Works for me :)

lukasbestle:
Thanks. 👍

PR #1999, opened 2023-04-07
Texnixe/meta docs

texnixe:


content/docs/1_guide/2_read-me/guide.txt
```diff
@@ -0,0 +1,251 @@
...
+----
+
+Intro: Guide, Reference, Cookbook, Search, where to find what and how to best use it
+
+----
+
+
+Text:
+
+Every user comes to a documentation with different expectations, background and prior knowledge. Ideally, Kirby or any other software/product would be self-explanatory, but that's a dream. It is equally a dream to make everyone happy, and consequently some of you will love this documentation and some of you will hate it, as our last survey clearly showed. But that's no excuse, and we will strive to improve it constantly.
```
lukasbestle: I feel like this paragraph is more for users who already experienced with Kirby (and don't need this guide page). New users might be discouraged by "some of you will hate it".

TBH I feel like we don't need this paragraph at all.

content/docs/1_guide/2_read-me/guide.txt
```diff
@@ -0,0 +1,250 @@
...
+
+## Parts of the documentation
+
+This documentation has <del>three</del> four parts:
+
+### 1. Guide:
+That's where you currently are and where you should start when you are new to Kirby to get an overview on the system. The Guide is divided into topics, and is the place where beginners should start.
```
lukasbestle: I think the second sentence is duplicated because the topics are mentioned below and the "beginners should start" part is mentioned directly before.

```suggestion
That's where you currently are and where you should start when you are new to Kirby to get an overview on the system.
```

lukasbestle (commented):
I committed a few minor changes directly. Hope that's OK.

General opinion:

I really like the the sections "Parts of the documentation", "A Reference page in detail" and "Parameters". Really good overview and easy to follow.

I'm not sure about the return types and exception sections though. If new users read these before other parts of the guide, they might get lost because some parts are quite advanced and require the understanding of general concepts in Kirby (e.g. what even is a field object). As a new user I'd be clicking around to all the methods you linked and be overwhelmed pretty quickly.

I think we could ease this with three changes:

- Remove the links to all the used methods. Rather explain in the info box at the top that the specifics of the used methods are not important yet and will be explained in later sections of the docs.
- The `Field` class is a pretty complex example when it comes to types (because it's also usable as a string and is returned from any method that doesn't exist). A simpler example for an object could be something like `$site->find('blog')` that returns a `Page` object.
- The void example with save and trigger requires the understanding of hooks. A simpler example could be `$kirby->user()->logout()`.

content/docs/1_guide/2_read-me/guide.txt
```diff
@@ -0,0 +1,250 @@
+Title: Read me first
```
lukasbestle: TBH I find it a bit weird that the title says "read me first" but this guide section is the second section after the quickstart guide.

The order makes sense to me (maybe even as third after the "Exploring the Starterkit" guide?). But the title could be something like "About the docs" to avoid a mental conflict between the quickstart and this guide.

lukasbestle (approved):

PR #2000, opened 2023-04-11
Texnixe/update image guide

texnixe:


distantnative:
It confused me quite a bit that the "Responsive images" guide entry isn't a guide but throws me into the cookbook. I think if we list it in the guide sidebar, it shouldn't take me out of the guide docs - that's quite disorienting. Same for "Image lazy‑loading".

texnixe:
Yes, I'm aware of that, but I'm not sure what would be a good solution. I wanted to give those recipes more visibility, but moving them to the guide doesn't seem to make sense. Maybe if we add an arrow to the links, it would make it clearer that those links lead out of the guide.

texnixe:
I solved this now with an explicit hint to the cookbook

lukasbestle:
I agree with Nico. IMO even with the `-> Cookbook` note it is still confusing that you are redirected to a completely different area of the site with no easy way to get back/navigate to the "next" recipe in the menu.

Also TBH I'm not sure about Files being a top-level guide category. IMO it still fits better under Content. Of course not if we need the subpages for the recipes and I guess this was the reason why you made it a top-level page.

texnixe:
In the first place, I made it a top-level page section to give it more attention rather than being buried under content, and yes, with those links to the recipes it was definitely necessary.

I don't really agree about this being confusing with the additional information, if I'm redirected from an in-page link, I have the same effect that I have to use the browsers back buttons.

Why not try it and see how users react to this?

distantnative:
I think this just shows again our main issue: we don't have a good structure for our Guide (anymore) and we keep putting it off. Putting another section on the top-level to get it more attention is no real cure for that. It's our go-to answer (e.g. page builder, UUIDs before) when we have no idea, but it isn't a solution actually. With 24 top-level sections with no sectioning between them we just give up guiding the user in any way and throw all at once at them. And I think that already has been voiced by our users.

I won't say no to the PR because of this - but I fear if we just keep merging them and not tackle the actual problem, it will just get worse and pop up again and again.


One other thing I noticed, Files in the Panel has a problem:
Screenshot 2023-04-16 at 20 30 43

lukasbestle:
I agree. To be honest I'm leaning more towards a better tree structure. We could allow grandchildren of guide sections. Then we could have subpages in the `Content > Files` guide.

Instead of the links to the Cookbook recipes, we could have short guide pages with the gist and a link to the respective recipe with more details and real-life examples.

texnixe:
I agree that we have a structure problem with the guide, but I'm not so sure about deeper nesting. I think I'd rather solve this with different sections.

I also don't agree that files should be under content, because it touches different parts, e.g. part of this was under templating, that's why I changed the structure. The resizing part was in the templates section, while the rest in content. This doesn't make sense from my perspective.

distantnative:
I also think it needs another level - and you both kinda say the same there. But it needs also a visual change. I also wouldn't see this level another layer deeper, harder to find, but rather our 24 top-level guides should be in 3-5 different sections to make them more approachable.

That combined with moving around probably quite a bunch of the guides.

texnixe:
I think if we structure the guide overview a bit better visually and in sections that are somehow related, that would do a big deal to making it more attractive, without having to introduce another nesting level (subheadlines, colors, whatever)

distantnative:
But putting the top-level items into sections is another nesting level (just not below but on top). That's what I meant.

texnixe:
Yes, ok, then we are on the same page, but that's a different type of nesting that putting more stuff in content...

texnixe:
I'm still for merging this, it's still an improvement over the current way of documenting the files stuff, and we cannot always wait for x (in this case restructure stuff) to happen first before we move on with something.

bastianallgeier:
Ok, I think we all agree that we need a different structure for the guides. I'm also not a fan of yet another nesting level. It's definitely personal taste, but when I look at the guide menu, I still think that the list is fine. It's sometimes easier to scan such a longer list than to navigate through each item and search for nested items. 

The bigger problem is the guides page and that's a design problem. I don't have a perfect solution right now and I feel like we should not let this fall behind again, but there's also not enough time right now to take care of it. 

I like the files guide in general and I would merge it like that. The only real issue I have immediately are the two cookbook links. They feel like a new paradigm for menu items that I rather would not introduce. 

What about a dedicated page with additional learning resources? Maybe "Resources", or "Further reading", or "Tutorials"? I think we could probably add more stuff there without filling up the menu.

texnixe:
I started with adding categories to the guides chapters yesterday, to better group them visually without introducing another nesting level. So the structure would remain the same, and there will just be subheadlines. That's a matter of finding the right grouping and the right headlines, though.

@bastianallgeier Well, if nobody likes those links in the menu, that would be a compromise, it's a step more though for users and those resources might be more easily overlooked. I would then update the PR.

lukasbestle (approved):

PR #2006, opened 2023-04-16
Add Remote class method examples

texnixe:
Remote defaults has the encoding value to be passed to `CURLOPT_ENCODING` set to `utf-8`. According to the curl documentation, valid values are null, empty string, gzip, br and identity. Is that a bug?


site/snippets/docs/default-params.php
distantnative, line 1: Can we give this a more precise name so that it's clear where it belongs to?

content/docs/3_reference/4_objects/http/0_remote/reference-class.txt
```diff
@@ -0,0 +1,36 @@
...
+Note that instantiating the object will fetch the data.
+
+You can also get an object by calling the `Remote::request()echo ` method or one of the shortcuts;
+
+
+### Shortcuts
+
+Instead of using `Remote::request()` with the request method defined in the `$params` array, you can use the HTTP method names as method:
+
+- `Remote::get()`, see (link: text: `$remote::get()`)
```
distantnative: is this missing a link value?

distantnative (approved):

PR #2020, opened 2023-05-07
Add categories to guide topics

texnixe:
This is just a first rough draft. Maybe we can find better category names/groups together. The same would then apply to the sidebar, similar to the Reference.

Of course, we could also style this in a nicer way.

### DIFF
--- a/site/templates/guides.php
+++ b/site/templates/guides.php
@@ -2,8 +2,13 @@
 
 <div class="mb-36">
   <h1 class="h1 mb-12">Guide</h1>
-  <ul class="guides auto-fill auto-rows-fr mb-12" style="--min: 16rem; --gap: var(--spacing-12)">
-    <?php foreach ($page->children()->listed() as $guide): ?>
+	<?php $categories = $page->children()->listed()->pluck('category', ',', true);
+	?>
+	<?php foreach ($categories as $category): ?>
+	<h2 class="h2 mb-12" style="font-size: var(--text-4xl)"><?= $category ?></h2>
+  <ul class="guides auto-fill auto-rows-fr mb-24" style="--min: 16rem; --gap: var(--spacing-12)">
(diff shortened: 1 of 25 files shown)


lukasbestle:
I quite like it already. A few thoughts:

- An alternative name for "Data & Frontend" could be "Structure & Semantics".
- Maybe we could rename the "Configuration" guide page to "Options". Then it will be "Configuration > Options" instead of "Configuration > Configuration".
- IMO the Routing, Caching and Emails guides don't really fit in Configuration. Maybe we could move these to a new "Going further" section together with Database.

lukasbestle:
Just saw that the icon for "Files" is currently the same as for "Content". Maybe we can find another icon for either to differentiate the guides.

site/templates/guides.php
```diff
@@ -2,8 +2,13 @@
 
 <div class="mb-36">
   <h1 class="h1 mb-12">Guide</h1>
-  <ul class="guides auto-fill auto-rows-fr mb-12" style="--min: 16rem; --gap: var(--spacing-12)">
-    <?php foreach ($page->children()->listed() as $guide): ?>
+	<?php $categories = $page->children()->listed()->pluck('category', ',', true);
```
distantnative, line 5: I'd move this to the controller

distantnative (commented):
Nice!

site/templates/guides.php
lukasbestle: Maybe we could also use `groupBy()` instead.

texnixe: We should, I guess I didn't because I originally had some guides that didn't have a category yet, which usually leads to an error.

texnixe:
> Just saw that the icon for "Files" is currently the same as for "Content". Maybe we can find another icon for either to differentiate the guides.

I know, there was no fitting icon left, we need a new one.

texnixe:
Closing because done

PR #2027, opened 2023-05-17
Update JUNO developer page slug to fix plugin update check in panel

SebastianEberlein-JUNO:
Our plugins use `Kirby::plugin('junohamburg/pluginname', [])` and the developer page slug was `getkirby.com/plugins/juno/pluginname`, so the update check in the panel did not work.


afbora:
Thanks 👍

PR #2036, opened 2023-06-01
Updated my profile name, avatar image

medienbaecker:
Inspired by other partners' profiles I updated my name that got truncated before. I also remove the white background from the avatar image.


medienbaecker:
Thanks for the quick merge, @texnixe

PR #2042, opened 2023-06-05
Add Typographer plugin

nilshoerrmann:
This PR adds our Typographer plugin to the directory.


texnixe (commented):
Could you please add the `Published` field with today's date? For some reason I cannot edit the file myself. Thanks!

nilshoerrmann:
Do the other plugins need that as well because I don't see that field there either?

texnixe:
We only added the field a while ago, doesn't make sense for old plugins. The purpose is to show new plugins for 6 weeks in the `new` category.

nilshoerrmann:
Hm, seems like something is not correct with my content file: the image is not displaying in the overview and there is no content on the plugin's page: https://getkirby.com/plugins/hananils/typographer

Any idea what I'd have to change to correct this?

texnixe:
File name is wrong, must be plugin.txt not default.txt

PR #2045, opened 2023-06-07
Partner filters

lukasbestle:
An initial draft for filters in the partner directory overview.

Completely without styling so far, I'd like to leave that part to Nico or Bastian. But the logic already works quite well I think. I've designed it to be easily extensible with additional filters if needed (just by adding more `data` attributes).

### DIFF
--- a/assets/css/framework/forms.css
+++ b/assets/css/framework/forms.css
@@ -1,0 +1,33 @@
+.select {
+	display: grid;
+	align-items: center;
+	grid-template-areas: "select";
+}
+
+.select select {
+	appearance: none;
+	background-color: transparent;
+	color: var(--btn-color);
+	border: 2px solid var(--btn-bg);
+	border-radius: 0;
+	font-family: var(--font-mono);
+	font-size: var(--text-sm);
+	line-height: inherit;
+	padding: 0.5rem 2rem 0.5rem 0.8rem;
+}
+
+/* Allows positioning of the chevron */
+.select select, .select::after {
+	grid-area: select;
+}
+
+/* Chevron */
+.select::after {
+	content: "";
+	background-color: var(--btn-bg);
+	clip-path: polygon(0 25%, 20% 25%, 50% 55%, 80% 25%, 100% 25%, 50% 75%);
+	height: 0.8rem;
+	width: 0.8rem;
+	justify-self: end;
+	margin-right: 0.7rem;
+}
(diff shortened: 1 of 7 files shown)


distantnative:
I like it in general!

However, I have quite a long delay when selecting "Turkey", which surprised me. Maybe it's actually faster just to hide all first, then show those that match again (instead of keepingg track of them in the object)?

And we also need to fix the spacing when no plus partner is matched.

I would also add Language as filter right from the start. That's maybe even more relevant than the country.

lukasbestle:
+1 on a language filter. I'm taking care of that.

I can't reproduce the delay with Turkey. The code is not keeping track of all partners in the object, but the other way around: Each partner keeps track of all filters it doesn't match. I think this is already the most efficient way as it scales well even with many partners.

Spacing: I thought about that as well, but decided against it for now because the needed code is IMO not worth it for the little effect (the margins collapse anyway, so it's only about the difference between `--spacing-36` and `--spacing-42`, which is 1.5rem). If we want to hide an entire section when all elements are hidden, we suddenly *do* need to keep track of all partners. We could still add this enhancement later if we find it important enough.

texnixe:
Great. I just wonder if such a fine-grained country filter currently makes sense, given that many combinations don't return any results. Or filter the options depending on what is already selected.

I can't reproduce the delay with Turkey either.

And also +1 for the language filter.

lukasbestle:
Added the language filter.

I think the country filter can be helpful as a secondary filter if there are too many results for the other two. We could hide it for now or we could make it a region filter instead (just continents).

texnixe:
I think a region filter would be more suitable at this point.

lukasbestle:
Changed it to a region filter. I also found a simple enough way to hide sections without visible children.

texnixe:
When filtering, results from the standard partners might be overlooked when there are also Plus partners, e.g. filtering by Dutch returns 2 results, but there seems to be only one unless you scroll down. Of course, that is how it is, but I think when filtering and there are only few results, that is a bit confusing. Maybe adding the number of results below the filters could alleviate this.

It might also make sense to add a sentence saying there are no results when there are none.

lukasbestle:
Agreed and done. :)

I think this only misses styling for the filters and the "no results" message now.

lukasbestle:
To revive this, I added basic styles that match the button styles:

Bildschirmfoto 2023-07-28 um 22 43 54

Here's a screenshot of the view when no results are found:

Bildschirmfoto 2023-07-28 um 22 44 26

Do you think it's mergeable like this?

texnixe:
I'd say yes, we can always change it later.

lukasbestle:
Updated the styling to work in Safari as well and made the line height consistent with buttons. Pretty happy with it now. :)

PR #2058, opened 2023-07-06
Update reference-component.txt

bnomei:
array might be null

### DIFF
--- a/content/docs/3_reference/7_plugins/3_components/0_url/reference-component.txt
+++ b/content/docs/3_reference/7_plugins/3_components/0_url/reference-component.txt
@@ -7,7 +7,7 @@ Text:
 ```php "/site/plugins/urls/index.php"
 Kirby::plugin('my/urls', [
   'components' => [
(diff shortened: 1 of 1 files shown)


content/docs/3_reference/7_plugins/3_components/0_url/reference-component.txt
```diff
@@ -7,7 +7,7 @@ Text:
 ```php "/site/plugins/urls/index.php"
 Kirby::plugin('my/urls', [
   'components' => [
-    'url' => function (Kirby $kirby, string $path, array $options = []): string {
+    'url' => function (Kirby $kirby, string $path, ?array $options = []): string {
```
afbora:
```suggestion
    'url' => function (Kirby $kirby, string|null $path = null, array|string|null $options = null): string {
```

Based on source: https://github.com/getkirby/kirby/blob/3.9.5/config/components.php#L331

bastianallgeier (approved):

PR #2063, opened 2023-07-12
Update PHPUnit cookbook recipe

ovenum:
Small update to make the recipe compatible with the latest PHPUnit version (10).
- Change link to PHPUnit getting started page to the latest version.
- Test File has to have the same name as the test class defined inside.


bastianallgeier (approved):

bastianallgeier:
Thank you for the changes!

PR #2071, opened 2023-07-31
fix: more security config in nginx article

pReya:
- Include headers to avoid MIME sniffing, just [like the .htaccess file](https://github.com/getkirby/starterkit/blob/main/.htaccess#L62-L65)
- Allow webserver to serve files from the `.well-known` folder, [just like the .htaccess file](https://github.com/getkirby/starterkit/blob/main/.htaccess#L27)
- Include disclaimer about boilerplate character of the given config file


lukasbestle (approved):

lukasbestle:
Thank you 💛

PR #2072, opened 2023-08-02
Update layoutMethods reference-extension

nerdcel:
Use $this instead of $layout in the example code


bastianallgeier (approved):

bastianallgeier:
Thanks for the fix

PR #2077, opened 2023-08-18
Adds move permission to files and pages

tobiasfabian:


lukasbestle:
Thanks for your PR, however we cannot merge it yet as the permission was introduced in v4, which is not yet stable.

I've added it to our [v4 docs](https://github.com/getkirby/getkirby.com/commit/549a7e0eb7490e07428299e050b45cb3084a46e0) so we will not forget about it.

PR #2094, opened 2023-09-22
Update reference-classmethod.txt

nilsmielke:
The syntax `$results = $page->search('my awesome search', 'title|text');` throws the error `Kirby\Cms\Page::search(): Argument #2 ($params) must be of type array, string given, called in {…}` . Needs to be `$results = $page->search('my awesome search', [ 'fields => [ 'title', 'text' ]]);` to succeed.

This is in Kirby 4.0.0-beta.1


distantnative:
@nilsmielke Isn't this than rather a bug in v4?

We also shouldn't make v4 changes to the docs before v4 has been released.

nilsmielke:
@distantnative Probably true. Unfortunately I do not have search set up in a k3 site right now.
Is this rather an issue to be reported in the dicord v4 channel, then?

lukasbestle:
It's indeed a regression in v4. I've opened a core PR here: https://github.com/getkirby/kirby/pull/5675

nilsmielke:
Thank you, @lukasbestle.

PR #2110, opened 2023-11-05
Use Panel dist icons

distantnative:
Not sure if this could be an issue in terms of performance or just fine.
(or maybe one could use a static variable to store the content of the list SVG file to not load it multiple times in one request?)

But I think it would be great to use the Panel icons and, hence, have more consistency between getkirby.com and the Panel.


site/plugins/site/helpers/html.php
```diff
@@ -13,7 +13,17 @@ function ariaCurrent(
 
 function icon(string $name): string|false
 {
-	return svg('assets/icons/' . $name . '.svg');
+	if ($svg = svg('assets/icons/' . $name . '.svg')) {
+		return $svg;
+	}
+
+	if ($panel = svg('kirby/panel/dist/img/icons.svg')) {
```
lukasbestle: Caching could be as simple as:

```suggestion
	static $panel;
	$panel ??= svg('kirby/panel/dist/img/icons.svg');
	if ($panel) {
```

lukasbestle (commented):
I like the idea to take the icons from the Panel set, avoids duplication. Haven't tested though whether they work visually and fit with the remaining custom icons.

distantnative:
@lukasbestle the style isn't an issue since we already replaced almost all getkirby.com icons with remix icons on the v4 branch. So the getkirby.com icons are already the same as the ones we would use from the Panel dist

site/plugins/site/helpers/html.php
distantnative: Thanks, added!

assets/css/site/logo.css
```diff
@@ -4,6 +4,8 @@
 }
 .logo svg {
   display: flex;
+	width: 36px;
```
distantnative, line 7: @bastianallgeier alternatively to setting all svg to that size, we could wrap all those inserted via `icon()` helper with a span that sets this size.

bastianallgeier:
TBH, I would leave this for after the launch. It feels like unnecessary stress to get this done.

bastianallgeier (approved):
I'm happy :)

PR #2115, opened 2023-11-09
Filter by compatible version

afbora:
I've implemented rough version. I realize there are many shortcomings. I'll proceed according to your reviews. Here few details:

- Filters multiple versions (3.5, 3.6, 3.7, 3.8, 3.9, 4.0 not only 3 or 4)
- Checks based on `Versions` content value that constraint syntax `^3.8 || ^4.0`
- Shows compatible versions on the plugin cards

screen-capture-1734-Plugins - Kirby CMS-getkirby test

### DIFF
--- a/assets/css/site/filters.css
+++ b/assets/css/site/filters.css
@@ -6,6 +6,7 @@
 }
 .filters svg {
   margin-right: var(--spacing-3);
+	width: 18px;
 }
 .filters hr {
   margin: var(--spacing-3) 0;
--- a/assets/css/site/menu.css
+++ b/assets/css/site/menu.css
@@ -165,3 +165,12 @@
 		margin: 0 0.75rem;
 	}
 }
+
+/* Versions */
+li.compatibility a {
+	display: inline-block;
+}
+
+li.compatibility a:not(:last-child)::after {
+	content: ',';
+}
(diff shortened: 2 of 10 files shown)


bastianallgeier:
This is a great start. I think we should keep this as simple as possible. I already love the tags for each card and we should definitely move this forward. I would just add a single menu item for v4 plugins though and ignore all the v3 subversions for now. We can still do this later, but for the launch we need to keep it simple. Can you develop this further @afbora and also look at a way to get the compatibility info for as many plugins as possible? This would be a HUGE help.

afbora:
> also look at a way to get the compatibility info for as many plugins as possible? This would be a HUGE help.

I can move version information here while I try/test the plugins. But I'm waiting for RC1 so I don't go back again 🙂 

> I would just add a single menu item for v4 plugins though and ignore all the v3 subversions for now

When we already have the compatible version information of the plugins, v3s will also work. I don't think it will bring an extra big workload.

lukasbestle:
> When we already have the compatible version information of the plugins, v3s will also work. I don't think it will bring an extra big workload.

I think it makes sense to already collect the compatibility information also for 3.x where we know it. But I wouldn't invest time actually testing with 3.x right now. So it is IMO likely that not all plugins will have compatibility info for all versions. So v3 filters aren't really useful as they won't show all relevant plugins, no? Also I think v3 filters would add too much visual clutter at the moment when people are mostly interested in v4 plugins. I agree with Bastian to keep it as simple as possible for now regarding the functionality. As I wrote, the collected data can still be more specific than what we provide as filters, but it really doesn't have to be.

PR #2118, opened 2023-11-10
Buy plugin

distantnative:
- [x] Merge first https://github.com/getkirby/getkirby.com/pull/2117


site/templates/buy.php
```diff
@@ -295,18 +307,25 @@ class Price extends HTMLElement {
...
 
-document.addEventListener("click", (event) => {
-	[...document.querySelectorAll("details")].forEach(details => {
-		if (details.contains(event.target) === false) {
-			details.removeAttribute("open");
+	for (const link of [...document.querySelectorAll(".checkout-link")]) {
+		link.href += window.currency;
+	}
+
+	document.addEventListener("click", (event) => {
```
lukasbestle: Why was this event listener moved inside the Paddle callback?

site/plugins/buy/Product.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+		return option('buy.' . $this->value . '.id');
+    }
+
+	/**
+	 * Returns the price object for the product
+	 */
+	public function price(string $currency = 'EUR'): Price
+    {
+        return new Price($this , $currency);
```
lukasbestle, line 58: I think it would be good to have another fallback to `EUR` for the currency when the passed value is an empty string. This will happen when the Paddle script in the frontend didn't execute for whatever reason and therefore the links couldn't be updated with the user currency. With the fallback to `EUR` we will at least have a working checkout.

site/config/buy.php
```diff
@@ -0,0 +1,22 @@
...
+		'nudge'   => 59,
+		'regular' => 99,
+	],
+  	'enterprise' => [
+		'id'      => 824340,
+		'min'     => 199,
+		'nudge'   => 249,
+		'regular' => 399,
+	],
+	'sale' => false,
```
lukasbestle: So far we have the ability to schedule sales via the Paddle backend and also our banner code on the site. I think it would be good to be able to do the same with the actual sale pricing, so basically set a start and end timestamp here and the sale will only be active during this time.

site/plugins/buy/Price.php
```diff
@@ -0,0 +1,97 @@
...
+
+	/**
+	 * Gets the price for a single license
+	 * at the given volume
+	 */
+	public function volume(int $volume): float
+	{
+		$price    = $this->sale();
+		$discount = option('buy.volume')[$volume];
+		$price   *= (1 - ($discount / 100));
```
lukasbestle: With something like this we could support arbitrary volume package sizes:

```suggestion

		// sort discounts by largest package first
		$discounts = option('buy.volume');
		krsort($discounts);

		// find highest possible discount for given volume
		$discount = 0;
		foreach ($discounts as $package => $packageDiscount) {
			if ($package <= $volume) {
				$discount = $packageDiscount;
				break;
			}
		}

		$price *= 1 - $discount / 100;
```

site/plugins/buy/rates.json
```diff
@@ -0,0 +1,31 @@
...
+	"PLN": 4.449,
+	"SEK": 11.634,
+	"SGD": 1.4523,
+	"THB": 38.089,
+	"TWD": 33.42,
+	"TRY": 30.4699,
+	"UAH": 38.54,
+	"USD": 1.0691,
+	"ZAR": 19.8926
+}
```
lukasbestle, line 31: Do we have a script to generate those rates (also including purchasing power parity discounts)?

site/plugins/buy/Price.php
```diff
@@ -0,0 +1,97 @@
...
+class Price
+{
+	public readonly float $rate;
+
+	public function __construct(
+		public readonly Product $product,
+		public readonly string $currency = 'EUR'
+	) {
+		$rates = Data::read(__DIR__ . '/rates.json');
+		$this->rate = $rates[$currency];
```
lukasbestle: This will explode for an invalid currency.

distantnative: The JSON entries and the list of Paddle currencies need to be  the same anyways, so that's not a case that needs to be supported.

site/plugins/buy/rates.json
distantnative: No, we said to solve that at a later point

site/plugins/buy/Price.php
distantnative: Is this something we want? We haven't offered that so far.

Looks good if we think we want to allow for that.

site/templates/buy.php
distantnative: I think purely a merge conflict as Basti was implementing the details while this code already was living on the other branch

site/config/buy.php
distantnative: So basically to add a check in `Price::sale()` if `buy.sale` is an array. And if so, check if current date time is between `start` and `end` entries of that array.

site/plugins/buy/Price.php
lukasbestle: We already had higher package sizes in the array before, I think for that purpose.

distantnative: Added it but using `A::find()`

site/plugins/buy/Price.php
distantnative: Added a fallback to EUR that also changes sets `currency` to EUR then

site/plugins/buy/Product.php
distantnative: Added

distantnative:
@lukasbestle would be great to get your knowledge to check if I transferred the expires logic from banner helper to Sale class correctly. 🙏

bastianallgeier (approved):
I would suggest to merge this already and add fixes later if needed

lukasbestle:
@distantnative Reviewed the sale changes. Very clean code, I like it. And we really didn't need the complexity of multiple banners.

I have made fixes to the date calculation, but that was already broken the same way in the old version. I've directly committed the fixes to the v4 branch.

PR #2119, opened 2023-11-13
Filter plugins by compatible version

bastianallgeier:
This is another shot at a more simplified plugin filter after our conversations on Discord. 

It uses @afbora suggested version bubbles, but keeps the colors out so far. I've also added a simple compatibility display to the plugin profiles. 

To make this work, I had to add the Versions field with a pre-filled 3 to all plugins. I hope this won't clash too much with whatever change happens on main. We could also think about running this on main already to avoid such clashes.


lukasbestle:
Can we have `title` attributes for the labels with something like "This plugin supports Kirby 3"? Just `K3` is hard to understand, especially to screenreader users.

bastianallgeier:
@lukasbestle good idea! I've just added them.

distantnative (approved):
I changed the version box/button on the plugin template a little. Before it sounded too much like the plugin's version to me.

PR #2120, opened 2023-11-15
Update release-guide.txt

luxuryluke:
code indenting doesn't copy correctly from docs page.
I don't think this fixes it, though.


lukasbestle:
Thanks, it was mixed spaces and tabs. YAML only supports spaces for indentation.

luxuryluke:
Bane of my yaml dev life. 😆

PR #2122, opened 2023-11-18
chore: changed plugin descriptions with hints for k4

mauricerenck:


lukasbestle:
We will have a new filter for Kirby 4 plugins that uses a new field like this:

```
Versions: 3, 4
```

I think it makes sense to already use this field instead of the text additions, then the filter and version bubbles will work out of the box on the Kirby 4 release.

mauricerenck:
That sounds good, thanks for the hint! I changed the files accordingly

PR #2131, opened 2023-11-28
Fix security.json markup

distantnative:


lukasbestle (approved):

lukasbestle:
Thank you 💛
LGTM, let's test it.

lukasbestle:
JSON looks perfectly now and the update check should fix itself after all caches (KeyCDN and local update check cache) are updated.

However the HTML template at https://getkirby.com/security now throws a fatal error.

PR #2133, opened 2023-11-28
Update guide.txt - remove 'form'

luxuryluke:
Technically they are 'form' fields, but for a person reading this page, it should be specifically referencing a 'blueprint' field.

### DIFF
--- a/content/docs/1_guide/7_blueprints/2_fields/guide.txt
+++ b/content/docs/1_guide/7_blueprints/2_fields/guide.txt
@@ -119,7 +119,7 @@ text:
 
 ## Field shortcuts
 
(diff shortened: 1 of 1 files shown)


texnixe:
Hm, then maybe better just call them fields, we don't talk about blueprint fields anywhere else.

luxuryluke:
Perfect.

Edit: Actually, I see **blueprints** mentioned on this page a few times.

texnixe:
Hm, this PR still has the term "blueprint fields", could you please just change to "fields"? Thank you!

luxuryluke:
There we go. Let's try this version.

PR #2136, opened 2023-11-29
Remove plugin and profile from directory

sebastiangreger:
Commentions plugin retired as incompatible with K4, due to unfortunate lack of resources :cry:


bastianallgeier:
@sebastiangreger that's really sad but understandable. Thank you for the PR.

PR #2137, opened 2023-11-29
Update content clean-up recipe

distantnative:
- Adds `focus` as ignore 
- Makes use of `$kirby->models()`

### DIFF
--- a/content/docs/2_cookbook/7_extensions/0_content-file-cleanup/cookbook-recipe.txt
+++ b/content/docs/2_cookbook/7_extensions/0_content-file-cleanup/cookbook-recipe.txt
@@ -6,11 +6,9 @@ Text:
 
 Kirby doesn't delete fields from content files if you later remove them from your blueprints, rename them, or if you add fields manually with an editor. Especially during development, you therefore often end up with a lot of unused stuff you will probably want to clean up before handing a project over to your client.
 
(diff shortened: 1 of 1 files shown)


content/docs/2_cookbook/7_extensions/0_content-file-cleanup/cookbook-recipe.txt
```diff
@@ -40,61 +40,61 @@ $kirby = new Kirby;
 $kirby->impersonate('kirby');
 
 // Define your collection
-// Don't use `$site->index()` for thousands of pages
-$collection = $kirby->site()->index();
+$models = $kirby->site()->index(true);
```
texnixe: Don't really understand the use of the `$models` variable name. As if I was going to loop through different model types like pages, users and files at the same time.

distantnative: Good call, forgot to change the right side to `$kirby->models()` which does that exactly

distantnative: Changed it.

I removed the paragraph about large index as `$kirby->models()` is a greater and yields each model. To my understanding, this should work also for larger sites.

PR #2138, opened 2023-11-30
Buy plugin refactoring

lukasbestle:


lukasbestle:
I‘ll work on this some more to use lazy-loading for the prices (increases perceived performance as it avoids the initial delay when loading the buy page).

PR #2145, opened 2023-12-04
Update reference-article.txt

luxuryluke:
Looks like odd yaml spacing, is all. Worried it might code copy incorrectly.


lukasbestle (approved):
Thanks! In fact YAML doesn't support tabs for indentation at all. I've fixed it for all examples on the page.

luxuryluke:
Oh, man, i fell for my own trap. used a tab, looked great and all lined up, then filed the PR. Oops. Great catch.

PR #2146, opened 2023-12-04
Update reference-article.txt

luxuryluke:
I'm guessing this part is missing for this portfolio setup.
Missing the file/cover.yml blueprint for cover images.

Of course, modify this to fit what you'd really like to say.

### DIFF
--- a/content/docs/3_reference/3_panel/8_samples/0_portfolio/reference-article.txt
+++ b/content/docs/3_reference/3_panel/8_samples/0_portfolio/reference-article.txt
@@ -26,6 +26,14 @@ listed:
   layout: cards
 ```
 
+### File cover blueprint
+
+We'll create a `cover` blueprint file for the project to reference a certain image as being more special than just an image. This will be useful since we'll be able to specify what image will represent the project.
+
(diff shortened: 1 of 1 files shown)


lukasbestle (approved):
Makes sense, thanks!

luxuryluke:
However, i did not test this.

texnixe:
It will actually work as expected even without that file, so to assign a file template, that template doesn't have to exist.

lukasbestle:
@texnixe Is that also the case when changing the file template in v4? I thought that only works with existing blueprints, but I may misremember.

luxuryluke:
Alternatively you could strip the Cover functionality out of this demo and it may be more straightforward and simple.
I loove the cover functionality myself.

texnixe:
@lukasbestle Good point, haven't checked.

luxuryluke:
Also, it mentions one MIN and three MAX cover images here, and that seems odd. Why is Cover used instead of Image, then? Preferences, clarity, etc.
Perhaps a bit of this portfolio code is worth simplifying or revisiting. No pressure at all, though.

texnixe:
@luxuryluke I agree.

PR #2161, opened 2023-12-26
Added explanatory info to Controllers guide TL;DR section

joh-sch:
I'd like to propose adding a bullet point to the TL;DR list on the _Controllers_ guide page, explaining that all data returned by a controller is available everywhere in the template it corresponds to. I believe this is a key aspect of the practical use of controllers and helps conceptually connecting controllers and templates.


lukasbestle (approved):
Thank you, that's indeed a very valuable piece of information.

PR #2167, opened 2024-01-05
Replace table field plugin

bogdancondorachi:
**This PR contains updates on the new table field plugin and a request to remove the old one from the directory, having the new one as a replacement.**

Table Field plugin updates:

1. Added license
2. Updated card image
3. K3 support

Also, a request to remove the old table field plugin made by @ragi96 due to lack of maintenance and the existence of an alternative that's based on his work. The new plugin offers migration compatibility with the old plugin, using the same content structure. Would be best to not endorse the old plugin in the directory so it does not mislead the users.


bogdancondorachi:
Also, the latest version does not seem to appear correct though the plugin directory shows the correct latest version. 😕 

image

afbora:
Latest version seems 2.1.0 in the directory:
https://getkirby.com/plugins/bogdancondorachi/table-field

And I just wonder that why you removed the old table plugin of ragi96?

bogdancondorachi:
> Latest version seems 2.1.0 in the directory: https://getkirby.com/plugins/bogdancondorachi/table-field
> 
> And I just wonder that why you removed the old table plugin of ragi96?

@afbora I know, but could be related to the same issue [reported](https://forum.getkirby.com/t/plugin-version-in-panel/30396) in the forum?

In my view, the current iteration represents an evolution of the ragi96 plugin, building upon its foundational concept. This version encompasses all the functionalities of its predecessor while introducing refinements and bug fixes to ensure seamless operation. Moreover, it prioritizes migration compatibility for a smoother transition. Given its unmaintained status, coupled with the availability of a stable alternative, it seems right to not endorse it within the directory to prevent potential confusion.

Of course, it's my opinion and it's totally up to the team. But this discussion can come again in a year or so when there could be more unmaintained plugins, take the WP directory as an example.

I can revert it if decide otherwise. 😃

distantnative:
I would say this makes sense in this case. Arguing functionalities, superiority etc. can always be quite subjective, so I'd not feel so comfortable to just decide it based on that without hearing ragi96's view on this. But the fact that it hasn't been maintained in the last 1.5 years and only support Kirby 3 I think make a good case to remove it in favor of this plugin here. Also not to confuse/mislead users.

Just would be great to make it clearer in the initial post, especially since it's altering/removing a plugin form someone else. Deletions of whole files are easy to overlook in the Github review UI, so I think it's good to be very transparent from the start :)

afbora:
Pinging @ragi96 to get his opinions.

ragi96:
I am completely fine with replacing it as I am atm quite far away from kirby and also don't have time to maintain it properly.

afbora (approved):

PR #2178, opened 2024-01-22
Less aggressive icon style

bastianallgeier:
If we only set the default size for those svgs that actually have the viewport definition of the remix icons, we should run into way fewer issues.

### DIFF
--- a/assets/css/site/icons.css
+++ b/assets/css/site/icons.css
@@ -1,5 +1,8 @@
 :where(svg) {
 	fill: currentColor;
+}
+
+:where(svg[viewBox="0 0 24 24"]) {
 	width: 18px;
 	height: 18px;
 }


lukasbestle:
I think it's a good fix. Can't imagine we use other SVGs that happen to have that viewBox. Don't we also need to do the same with the fill though? Illustrations, logos etc. shouldn't be recolored.

bastianallgeier:
As far as I can see, we are using the currentColor trick in other parts as well. At least for the Algolia icon. I think I would leave it like that and maybe adjust later if needed. It seems like we had only issues with the sizes, but not with fill.

lukasbestle:
Makes sense, basically all of our SVGs are not colorful anyway and on purpose.

PR #2187, opened 2024-01-31
Add embedded image example

kevinvennitti:


content/docs/1_guide/16_emails/guide.txt
```diff
@@ -277,6 +277,8 @@ $kirby->email([
     // …
 		'template' => 'email',
     'beforeSend' => function ($mailer) {
+				// $image should return file path
+				$image = asset('assets/images/logo.png')->root();
         $mailer->AddEmbeddedImage($image, 'image', 'image.png', 'base64', 'image/png');
```
bastianallgeier: I think that's a good addition, but then it could even make use of more asset methods. 

```suggestion
				// $image should return file path
				$image = asset('assets/images/logo.png')->root();
        $mailer->AddEmbeddedImage($image, 'image', $image->filename(), 'base64', $image->mime());
```

kevinvennitti: Indeed!

content/docs/1_guide/16_emails/guide.txt
```diff
@@ -265,6 +265,27 @@ $kirby->email([
...
+<!-- Email content -->
+```
+
+```php
+$kirby->email([
+    // …
+		'template' => 'email',
+    'beforeSend' => function ($mailer) {
+				// $image should return file path
+				$image = asset('assets/images/logo.png')->root();
```
texnixe: This looks wrong, $image in this line returns a string, then you cannot call `$image->filename()`. This should probably read:

```
$image = asset('assets/images/logo.png');
$mailer->AddEmbeddedImage($image->root(), 'image', $image->filename(), 'base64', $image->mime());
```

kevinvennitti: Yep, my bad. Next commit has fixed that 👍

PR #2191, opened 2024-02-04
New `archived` label for the plugins

afbora:
- Used a date for archived field instead bool. So we can hide or remove some archived plugins that that have been around for a long time.
- Feel free to update the style since I'm really bad on UI :)

screen-capture-1827-Entity Field - Kirby CMS-getkirby test

screen-capture-1828-Oblik Studio - Kirby CMS-getkirby test

### DIFF
--- a/assets/css/framework/buttons.css
+++ b/assets/css/framework/buttons.css
@@ -43,3 +43,7 @@ button {
 .btn--filled svg {
 	color: var(--color-yellow-600);
 }
+.btn--yellow {
+	background: var(--color-yellow-400) !important;
+	color: var(--color-black); !important;
+}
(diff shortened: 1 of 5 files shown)


bastianallgeier:
This is a great start. I think it's important to already filter out the archived plugins from the other categories though. But that would be a simple step. I'll try to add a few design tweaks today.

bastianallgeier:
I made a few small changes. 

- I moved the btn--yellow to the plugin template because we only need it there
- I removed the badge from the cardlet again. I don't think we need it there
- I've refactored the plugins controller and filtered out the archived plugins.

bastianallgeier (approved):

PR #2192, opened 2024-02-04
Add DDEV cookbook recipe

texnixe:


bastianallgeier:
Wow, that's very extensive. Super cool! I fixed a small indentation issue, but otherwise I'm happy. What do you think after the setup. Should we invest more time in a custom quickstarter? It almost seems to me like it's not really needed after all. The only main reason would probably be to be listed on their site.

bastianallgeier (approved):

distantnative:
> The only main reason would probably be to be listed on their site. 

Which I think on it's own would be pretty good if writing the quickstarter isn't too hard

texnixe:
Not quite sure what you mean with a custom quickstarter. I prepared something to put on their site, along the lines of this recipe, but shorter. Don't really think we need a custom kirby project type, looked into it but couldn't quite figure out how to test it locally.

distantnative:
@texnixe I think we mean having a section here https://ddev.readthedocs.io/en/latest/users/quickstart/

Whether custom project type or just short description, whatever works for DDEV and is easier for us.

texnixe:
Ok, with the short description we are ready to go, will push to my fork tonight, then we can discuss before creating a PR. With the custom product type, I couldn't figure out from the docs how to build it locally, therefore I think it makes more sense to go ahead with what we have to be listed there. We can always change that later should we decide that a custom project type would be useful. With the little setup it needs to get it running with Kirby, I don't really see a reason.

texnixe:
Bildschirmfoto 2024-02-05 um 09 35 02

bastianallgeier:
We could probably think about adding the CLI by default to the quickstarter later or other things like that, but the instructions are a great first step and I guess it also gives us less stuff to worry about.

texnixe:
Thought about the CLI as well and a custom command for that, although not quite sure how much sense that makes if the CLI is not in the Kirby core.

Setting up a ddev project via a custom CLI command would be an option as well.

bastianallgeier:
Let's go without the custom type. We can still extend any of this later if we want.

PR #2195, opened 2024-02-06
Re-showcase

afbora:
- 47 new cases added
- 4 cases removed
- 7 cases screenshots updated
- ~Showing case limited as 30~
- Screenshots doesn't optimized (FYI)

Feel free to add/delete/update cases. If you let me know, I can do for you 👍


distantnative:
Ahh this will be cool. Some quick thoughts:

- Any new cases that should go into https://github.com/getkirby/getkirby.com/blob/main/site/collections/causes.php ?
- Was thinking about adding a link to kirbysites already on the top, right of the headline. @bastianallgeier thoughts?
- Do we need an indication somehow that there are more and that one should reload to see a new shuffled set?

texnixe:
> Do we need an indication somehow that there are more and that one should reload to see a new shuffled set?

That would actually annoy me, if you only see a new shuffled set, with most likely many of the cases you have already seen, I think a load more button would make more sense. Or we show all, but with lazy-loading

afbora:
> Any new cases that should go into https://github.com/getkirby/getkirby.com/blob/main/site/collections/causes.php ?

I missed that collection. Let me check again.

> That would actually annoy me, if you only see a new shuffled set, with most likely many of the cases you have already seen, I think a load more button would make more sense. Or we show all, but with lazy-loading

Would lazy-load or load more button make sense?

If there will be a load more button, shuffle may be difficult. Because after shuffling and showing the first part, it is necessary to specify which ones should not be included in the second part. I think it is not worth it.

afbora:
@distantnative Here possible causes in new ones but not sure, please check again for me.

- https://klimadashboard.org
- https://researchstories.net/en
- https://scrollnichtweg.de/
- https://eunic.eu/
- https://evensfoundation.be/

texnixe:
> If there will be a load more button, shuffle may be difficult. Because after shuffling and showing the first part, it is necessary to specify which ones should not be included in the second part. I think it is not worth it.

@afbora Maybe lazy-loading is the better option. However, I don't see the problem. A shuffled collection is just collection that can be paginated. So no need to specify which should be included in the second part.

afbora:
> A shuffled collection is just collection that can be paginated.

Of course it can be paginated, but when you go to the second page the whole collection is re-shuffled (since each request is handled separately) and gives you the second page. In this case you need to cache the order of the first collection you randomly shuffled.

bastianallgeier:
This is fantastic and the new cases look really good. 

My suggestions (already added, feel free to revert)

1. I would also move the footer up and simplify it a bit …
Screenshot 2024-02-07 at 12 39 36

2. I would show all cases. Yes, it loads a bit longer, there are a lot of images and we really need to fix optimization via KeyCDN, but I think the list still has a length that is nice. As a visitor, I really want to scroll through lots of examples here and get inspired. So I think it's better to show all at once.

distantnative:
I think for https://github.com/getkirby/getkirby.com/pull/2195#issuecomment-1930630730 @bastianallgeier might now or we ended to check history.

bastianallgeier:
@distantnative yep, I forget to comment on that. I think unfortunately most of them have a license and we didn't support them – because they didn't ask. Otherwise they would be excellent.

distantnative:
I like it!

bastianallgeier (approved):

bastianallgeier:
I've merged it in this state and created a new project for the good causes update.

PR #2197, opened 2024-02-07
docs: Minor improvements to discussion of public folders and private folders

rfay:
In https://github.com/ddev/ddev/pull/5791 we're discussion the DDEV quickstart and Kirby "public folders". 

I noted that the upstream Kirby docs here might not be as clear as they could be, so here's a minor starter suggestion of what I'd encourage.

I do suspect that "public folder" isn't the right name for this class of setup. 

You're welcome to do as you will with this, including pushing any changes you want into it. I don't know enough about Kirby to offer any nuance

But we do very much welcome you to DDEV!

### DIFF
--- a/content/docs/1_guide/12_configuration/guide.txt
+++ b/content/docs/1_guide/12_configuration/guide.txt
@@ -233,11 +233,11 @@ echo $kirby->render();
 
 (reference: system/roots)
 
-### Public folder setup
+### Public and private folder setup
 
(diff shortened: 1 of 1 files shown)


bastianallgeier:
@rfay thanks for the suggestions. I really like them, but I would leave it to @texnixe who wrote the article to give it a final look.

texnixe:
Thanks, @rfay, for the changes.

PR #2202, opened 2024-02-10
Add info for Microseasons plugin.

scottboms:
Adding a new plugin that renders the current Japanese micro season in Kirby templates.


texnixe:
I added it manually because of conflicts. Tip: Always pull latest changes into your fork first.

PR #2205, opened 2024-02-12
Add Semantic Markup plugin.

scottboms:
Adds new kirby-markup plugin which includes a collection of advanced semantic markup tags for advanced typesetting in textarea fields. Each tag also includes a custom textareaButton.


texnixe:
This branch has conflicts and cannot be merged.

PR #2206, opened 2024-02-13
Update block models docs

bezin:
Add hint to manual class loading for block models


content/docs/3_reference/7_plugins/1_extensions/0_block-models/reference-extension.txt
```diff
@@ -29,6 +29,30 @@ Kirby::plugin('my/blockModels', [
...
+class HeadingBlock extends Block
+{
+    public function toHtml(): string
+    {
+        return '<h1>Custom HTML for headings</h1>';
+    }
+}
+```
+
+```php "/site/plugins/index.php"
```
texnixe: Wrong path here, should be "/site/plugins/block-models/index.php"

texnixe:
Thanks!

PR #2207, opened 2024-02-13
Texnixe/guide changes

texnixe:


distantnative:
I really like diving directly into the guides. I think the sidebar grouping could be tweaked a bit to make the groups visually even tighter.

Also really like putting the screencasts on top.

I think the TOC is a bit out of place now. I would suggest we also limit image/videos to the narrower width of the text. Then we would have space to the right to add the TOC there as sticky element and could use it as on the release pages, highlighting on scroll where we are on the page.

We should also finally fix the sidebar menu on mobile (coming last after the guide is bad).

What do you thin about the following?
- Add a prev/next guide navigation at the end of each guide (as seen in many other docs, helping the user to just read on)
- Having a standard element (snippet, visual distinct box) for "Further reading" that pulls the links from a distinct content field. Makes it easy to add cross-links
- I think we should rework many of the titles of the guides to make them easier to understand at a glance in the menu.Would you find suggestions helpful or do you want to first work on this on your own?

texnixe:
- I like the idea of the prev-next navigation.
- Also agree with the further reading as a separate field
- Regarding the titles, not sure, the current titles have already been reworked from the original ones. Which ones do you find particularly confusing?

Regarding the ToC as a sticky element, don't know what example you are referring to. But not so sure that would actually work unless we make sure that each ToC entry is very short. Plus, where would the ToC go on mobile then? Also, especially any overview images would become too small, imo.

> could be tweaked a bit to make the groups visually even tighter.

I leave that to a designer

distantnative:
For putting the ToC to the right, e.g. how vite or vue do it: https://vitejs.dev/guide/
Screenshot 2024-02-18 at 09 09 24

Screenshot 2024-02-18 at 09 10 24

On mobile, both turn into dropdown menus similarly to how we do it in the reference:
Screenshot 2024-02-18 at 09 10 49

Regarding titles:
- "About the docs" seems a little disruptive in the flow from Get up and running to exploring the starter kit
- "Data & frontend" are very technical terms, maybe rather "Create your site"
- "Creating content", "Managing files"... are the verbs necessary? Could be more to the point with just "Content", "Files"...
- Same with "Configuring the Panel": since the user doesn't even know the Panel yet, why not just "Panel"
- We have a "Configuration" group that has "Configuring Kirby" only as third option. Shouldn't that go first? Or it needs a different title
- "Multi-language configuration" can be "Multi-language" only inside the configuration group
- "Panel users & permissions" => "Users & permissions"
- "Authentication & Sessions" sound very mighty, could deter less technical users even though it has something as important as which login types the Panel uses. Maybe "Login & Sessions" could be easier to grasp?
 - "The power of routes" => "Routing"
 - "Sending emails" => "Emails"
 - Group "Extending emails" starts with Guide with the same name, either renaming the group or the guide maybe to "Plugins"
 - "Databases" and "Emails" don't really feel at home in these groups, maybe move "email" to the same group as "databases" but that group could be rather called "Integrations"
 - "Beyond Kirby" is not really beyond, but talks about how to use Kirby as foundation to integrate other systems. The title doesn't say a lot what to expect in there. Maybe something like "Combine with other systems"?
 - I like "Maintenance" but shouldn't "Troubleshooting" then go into that group too?

Just my thoughts. For me, jumping directly into the guides and having them in groups, really makes me understand the guide more as a journey for the user from start to end. And that's where these thoughts plop up, what I think would help the journey to connect better from one step to the other.

texnixe:
Ok, I thought you were referring to a sticky nav in our site, because you mentioned release pages. But your example has the very same problem, with longer ToC entries being cut off because there is not enough space. Also, the navigation disappears quite quickly if you reduce the window size, which would be the same in our case.

We had all these short titles before, I agree that even the current ones are not always ideal, but verbs help users more than just throwing nouns at them. My opionion.

But we can discuss such things for ages.

However, to move forward, I think we should do this in steps, i.e. merge first when the design fits, then continue with the rest.

texnixe:
Hm, maybe the separate Further reading field is not such a good idea after all, as that removes the entry from ToC as well. And in the guide, this is not used very often, anyway.

distantnative:
@texnixe I think you are right. Let's rather make this many smaller steps. This is already great. I'll make some coding style suggestions and maybe @bastianallgeier can have a look at the design of the sidebar and/or prev-next. And let's merge it then and put any other changes etc. in new PRs.

site/snippets/layouts/prevnext.php
```diff
@@ -0,0 +1,23 @@
+<nav class="mb-24">
+	<?php $columnCount = $page->prevListed(page('docs/guide')->index()->listed()) === null ? 1 : 2; ?>
```
distantnative: I would suggest to move `page('docs/guide')->index()->listed()` outside of the snippet. This would make the snippet more universally usable and we wouldn't call `->index()` three times. So passing it as `$siblings` or so to the snippet.

site/templates/guide.php
```diff
@@ -2,8 +2,13 @@
...
-	'menu'  => page('docs/guide')->children()->listed(),
+	'title'         => 'Guide',
+	'link'          => '/docs/guide',
+	'menu'          => collection('guides'),
+	'hasCategories' => true,
 ]) ?>
 <?php endslot() ?>
+
+<?php slot('prevNext') ?>
+<?php snippet('layouts/prevnext') ?>
```
distantnative: Passing `['siblings' => page('docs/guide')->index()->listed()]` here

site/snippets/sidebar/menu-grouped.php
```diff
@@ -0,0 +1,34 @@
+<?php
+extract([
+	'open' => $open ?? false
+])
+?>
+<?php foreach ($menu->group('category') as $category => $items): ?>
```
distantnative, line 1: Any chance to combine `menu` and `menu-grouped` snippets to have less redundancy?

bastianallgeier:
This is all going in a really great direction! I've added a first set of design changes.

distantnative:
@bastianallgeier I do like prev-next styling, but the menu group headers don't work for me. They look way too much just like the actual entries

Screenshot 2024-02-21 at 18 26 37

I think they need a lighter color and should be smaller in size than the actual entries. If monospace font as Sonja had it, isn't an option.

distantnative:
@bastianallgeier I like the new style of the grouped sidebar, but it cannot be sticky as it is now. Otherwise one has to scroll always until the very end of every guide article if I want to see the rest of the sidebar menu.

And thus, I think we need a denser version also. On a 13 inch laptop, I only see 2/3 of the menu.

distantnative:
Or we need to collapse the whole groups

texnixe:
It would be good if we could merge this soon. I'm working on the cookbook restructuring and want to move parts to the guide. Also, links to the cookbook need to be updated after the restructuring, so to prevent too many merge conflicts, I'd like to have this merged first.

bastianallgeier:
Sorry for the delay. The checkout pretty much ate all my time last week. I would revert the sticky positioning, but I really don't think that the menu is too long like that. Check out the Vue guide menu as a comparison: https://vuejs.org/guide/introduction.html We could maybe make it a little bit denser, but I wouldn't stress this too much to be honest. I also personally prefer to have less stuff collapsed. If we hide the main categories as well, it gets really hard to get a good overview. We already collapse the sub menus.

bastianallgeier:
I've removed the borders under each group and made everything a tiny bit denser. But I don't think we can compress it even more than that. 

Maybe we could merge it like that and make more adjustments later?

distantnative:
Yes, the Vue guide has no issue with a long sidebar because its sidebar scrolls separately and is always accessible. Ours is either too long when sticky or out of reach when not sticky. My comment on making it dense was related to "as compact that we can keep sticky".

But I agree that we could merge it like this and try to fix/improve that problem later.

distantnative:
Midterm we might want to think of a full-width header (like in the Refence, Vuejs docs etc.) and use this on all pages with different variants for the main body (centered max-width for some basic pages, much wider for others, having space for three columns for guide section...).

distantnative:
@texnixe merge as you please/when you're ready

PR #2219, opened 2024-02-21
Showcase update

afbora:


site/collections/causes.php
```diff
@@ -8,6 +8,13 @@
...
 		'strassenfeger',
-		'teteenlair'
+		'teteenlair',
+		'klimadashboard',
+		'viseta',
+		'nowar',
+		'throughpositiveeyes',
+		'gobanyo',
+		'ppks',
+		'adunare'
```
afbora: What do you think about using new `Cause: true` field instead static collection?

````php
<?php
// site/collections/causes.php
return function () {
    return page('love')
        ->children()
        ->filterBy('cause', true);
};

distantnative: I would still use `$kirby->collection('cases')` as entry but filtering that by a field, sure, makes sense to me.

bastianallgeier (approved):

bastianallgeier:
Great job!

PR #2225, opened 2024-02-25
Replace dynamic property with passing a callback to the sort() method

ovenum:
As [mentioned](https://discord.com/channels/525634039965679616/525641819854471168/1210500537963774012) on the Discord Server, the Sorting Cookbook article currently makes use of _dynamic properties_ that will cause a [deprecation warning in PHP 8.2](https://stitcher.io/blog/deprecated-dynamic-properties-in-php-82). Support for dynamic properties is set to be [removed in PHP 9](https://stitcher.io/blog/new-in-php-82#deprecate-dynamic-properties-rfc).

This pull requests replaces the use of dynamic properties in the article with a callback function. 
I’m not quite sure if this is how the `sort()` method should be used and the guide would require more examples how to change sorting order.

### DIFF
--- a/content/docs/2_cookbook/2_content/0_sorting/cookbook-recipe.txt
+++ b/content/docs/2_cookbook/2_content/0_sorting/cookbook-recipe.txt
@@ -149,34 +149,29 @@ Now, instead of numbers, our sizes field contains sizes like `S, M, L, XL, XXL` 
 // fetch the products from the structure field
 $products = $page->products()->toStructure();
 
-// map an order field to each item of the collection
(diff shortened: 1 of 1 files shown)


distantnative:
Is this ready to be reviewed?

texnixe:
@distantnative We have a few other places where `map()` needs to be replaced, this is on my list.

https://github.com/getkirby/getkirby.com/issues/2229

bastianallgeier:
@texnixe Could we maybe still start to merge this PR for this article?

texnixe:
I have already fixed this in another PR now

PR #2226, opened 2024-02-25
Add multi-language setup advice for UUID

kevinvennitti:
Open to any change to make it clearer (another warning below the multilang code block instead?)


texnixe:
Not happy with the wording, what is `default language entry` supposed to mean? I guess this should read `translation array for the default language` or something along those lines.

But somehow this warning doesn't fit the example, anyway, as we would need a different route setup for a multilanguage context.

kevinvennitti:
Okay I agree 👍

PR #2235, opened 2024-03-01
`(tabs…)` KirbyText grouping

distantnative:
Screenshot 2024-03-01 at 21 36 17

```
(tabs…)
=== Tab title 1
Tab content 1

=== Tab title 2
Tab content 2

=== Tab title 3
Tab content 3
(…tabs)
```

@texnixe I chose not to use it in any actual guide yet as example, because you are working currently on them. So I would leave that to you, where/how to use this best. But added one example in the style guide to show how it works.

### DIFF
--- a/assets/css/index.css
+++ b/assets/css/index.css
@@ -37,8 +37,9 @@
 @import "site/screencast.css";
 @import "site/search.css";
 @import "site/sidebar.css";
-@import "site/types.css";
+@import "site/tabs.css";
 @import "site/toc.css";
+@import "site/types.css";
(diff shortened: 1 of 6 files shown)


texnixe (approved):

texnixe:
@distantnative I think we should merge this, then I can play around with it and see how it goes.

distantnative:
@bastianallgeier Let's do that. Then you can better tell if it needs more changes. @bastianallgeier can also still come in and tweak the styling, if needed.

PR #2236, opened 2024-03-02
Tweak docs font sizes and gaps

distantnative:
@bastianallgeier @texnixe 
Trying a simple fix for getting more compact docs.

I know that this could rub Basti in a wrong way regarding overall consistency across the whole site, but I think since that would require quite a bit of time, this seems to me like it could be a good compromise for effort and benefit.

### DIFF
--- a/assets/css/framework/headings.css
+++ b/assets/css/framework/headings.css
@@ -42,6 +42,6 @@
 .prose h4 code,
 .prose h5 code,
 .prose h6 code {
-	font-size: .875em;
+	font-size: 0.875em;
 	font-weight: 400;
 }
--- a/assets/css/framework/prose.css
+++ b/assets/css/framework/prose.css
@@ -161,7 +161,7 @@
 	display: inline-flex;
 	font-size: 0.825em;
 	max-width: 100%;
-	padding: 0 0.5em;
+	padding-inline: 0.375rem;
 	border-radius: 3px;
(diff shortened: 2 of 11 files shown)


texnixe:
@distantnative I like it.

distantnative:
Great! Lege See what @bastianallgeier thinks.

The new article.css file annoys me a bit with just one rule in there.

bastianallgeier:
I like the direction, but I really get nervous with the headline size changes. I don't think they are helping as much as they feel weird when jumping from other pages. I would rather focus some more on optimizing the prose css even more.

distantnative:
You mean the top headline? I think they are way too big in the docs and just take up screen space without adding much information ("Guide"). Also from most of the other pages the layout shifts anyways a lot to the docs, so I wouldn't see it as an issue. Or we decrease the title size also on those other pages.

bastianallgeier:
I think it's a good idea to simply reduce the font size everywhere to keep it consistent

distantnative:
@bastianallgeier I've globally tweaked the large text sizes instead now

bastianallgeier:
@distantnative great change! I went through most of the pages and it seems to work really well across the board. I think we need to keep adjusting intro blocks and other parts next, but those are really just minor details.

bastianallgeier (approved):

PR #2237, opened 2024-03-02
Restyle author card

distantnative:
Screenshot 2024-03-02 at 21 59 41

@texnixe if you have a version with a bit higher resolution of your image that would be great. The old file had a circular cutout and I only found this smaller one in another folder.


bastianallgeier (approved):

bastianallgeier:
I like it!

PR #2238, opened 2024-03-03
Guide: mobile sidebar

distantnative:
Very rough first draft, cookbook etc. broken with it. But wanted to first discuss this with you to see if it's worth heading further into this direction:

Screenshot 2024-03-03 at 19 23 26
Screenshot 2024-03-03 at 19 23 34


bastianallgeier:
I like the use of a native select box for it! It definitely simplifies this a lot and still has great UX because the selects are great to navigate on mobile.

bastianallgeier:
I think we should set the option for the current page in the select dropdown to selected

distantnative:
@bastianallgeier I had that but wasn't sure if it's then clear what the select does (and also repeating the same text right next to each other - select and h1)

bastianallgeier:
I've changed the styling a bit, but I'm really happy otherwise with this version!! 
Screenshot 2024-03-14 at 15 04 04

distantnative:
💛

Issue #2155, opened 2023-12-14
Improve documentation about routes in multi-lang context


texnixe:
- [ ] lang scope
- [ ] routes with and without language code
- [ ] setting language scope manually via site->visit etc.

PR #2240, opened 2024-03-04
Add example for language-agnostic routes

bezin:
This way it is clear that you can use the `*` placeholder without a page context and that the first parameter will always be the language code.

### DIFF
--- a/content/docs/1_guide/14_routing/guide.txt
+++ b/content/docs/1_guide/14_routing/guide.txt
@@ -270,6 +270,22 @@ return [
 ];
 ```
 
+If you want your route to match every language, you may use `*` as a placeholder. The langugae code will be passed to the action callback as first argument:
+
+```php
+return [
+    'routes' => [
+        [
+            'pattern' => '(:any)',
+            'language' => '*',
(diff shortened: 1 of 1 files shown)


afbora:
I just wonder that what's the difference with the following examples?
https://getkirby.com/docs/guide/routing#multi-language-setup__language-scope

bezin:
Yeah, maybe a sentence is another rather than providing an extra code snippet?

afbora:
Feel free to push your changes. I just wanted to say I'm confused about difference existing samples with yours.

texnixe:
Hm, I also think this sentence here is more confusing than clarifying. Also, the resulting $language parameter does not refer to a language code, but the language object. The whole language routing chapter needs an overhaul, there is already an issue for that: https://github.com/getkirby/getkirby.com/issues/2155, which I will take care of once guide/cookbook overhaul are done.

bezin:
Okay, I'll be patient then and wait for the overhauled docs. BTW I used `function(string $language, $otherArguments)` as action callback signature and the language object is type-casted to a string then.

PR #2248, opened 2024-03-11
Merge installation guides into one menu entry

distantnative:


bastianallgeier:
I really like this change. @texnixe what was your last opinion on this?

bastianallgeier (approved):

bastianallgeier:
@texnixe ping

texnixe:
🤷‍♀️ If you think that it is better

distantnative:
@texnixe what would be the argument to keep the installation docs from "Get up and running" and "Installation guide" separate? And "Exploring the Starterkit" in the middle.

texnixe:
@distantnative The Starterkit in the middle doesn't make sense at all. The reason why I think it should be separate is to 1. keep the "Get up and running" short (i.e. you can do this in 1 minute without having to read a complete guide with subchapters), and to have an installation guide in its own right (so someone who looks for installation guide can immediately see it in the sidebar).

distantnative:
The "Get up and running" guide itself isn't getting any longer with this PR (actually shorter with one more tabbed section).

My take is that the added clarity and compactness for the sidebar outweighs the benefit of being able to spot "installation" as word... I think people looking for such a guide just as well will click on "get up and running" - it really just means the same. Or we rename that to "Installation", idk. But having two top-level entries that serve exactly the same topic is not good IMO.

texnixe:
@distantnative We will not agree on this, so let's agree to disagree. 

The separation makes even more sense IMO when installation also contains update guides and the dev environment.

bastianallgeier:
I think I get both positions. 

Just thinking out loud: 

Preparing the stage
- Quickstart
- Installation
  - Local Dev Environment
  - Install via Git
  - Install via Composer
  - Install via CLI 
- Exploring the Starterkit
- About the docs

Maybe it's mostly about renaming the "Get up and running" point to let it less seem like a duplicate for installation.

texnixe:
@bastianallgeier Thought that as well

distantnative:
If be ok with new names, but honestly then Quickstart also should not start with "Installing Kirby" - it just makes no sense of having a top-level entry "Installation" and then the most essential section on "Installing Kirby" is not part of that entry but another top-level entry. There's just no way that makes sense for users. It's not about what impression the tile gives, it's that we keep the actual content on the same topic spread across different parts of the guide.

distantnative:
It becomes also very obvious in my opinion because "Installation" itself has no content but is an empty tool-level category redirecting. While Quickstart is just a top-level entry with no subpages at all. Why not to merge these I really cannot understand.

distantnative:
And the specific installation guides as subpages of Quickstart or whatever it's called would still directly be visible as the guide opens up on the Quickstart with its subpages expanded.

texnixe:
IMO, this discussion is not leading anywhere and just a waste of resources. I've made my point and will stop it here.

distantnative:
And sorry for the multi comment and being so passionately on this: But with that proposed structure we'd basically say that we rather want the few people who care to immediately find more advanced setups like Composer and Git to have a marginally faster UX vs. the majority of people wanting to do the normal file install clicking first into the wrong guide and having to find their way around to the other guide that's actually talking about the most common installation method.

lukasbestle:
I quite like the current structure with the separate Installation page for the reasons Sonja mentioned. But I also see the argument that there is currently a duplication between the Quickstart and Installation pages.

What if we split the Quickstart page into separate pages to give them more space and so they can be read in the right order:

**Preparing the stage**

- Requirements
- Installation & update (already exists, maybe we could merge in some parts of the Quickstart that are not already there)
- Initial setup (would get paragraphs for creating the first account and activating the license)
- Exploring the Starterkit
- About the docs
- Where to go from here (with only the parts that are not docs, e.g. forum, Discord, plugins, community page etc.; separate page allows for more details and maybe a more visual presentation)

distantnative:
Trashing this

PR #2249, opened 2024-03-11
Guide: move multilang intro to main guide

distantnative:
Use the main "Multi-language" guide entry for this content instead leaving this one empty and having a separate introduction page


bastianallgeier:
That's definitely better.

PR #2250, opened 2024-03-11
Guide: move content intro to main guide

distantnative:
Use the main "Content" guide entry for this content instead leaving this one empty and having a separate introduction/overview page


bastianallgeier:
Same here! I think it makes sense to use the main points for it.

PR #2257, opened 2024-03-11
Partners: switch to `Info` field for clients

distantnative:


bastianallgeier (approved):
I'm happy with this but I think it's important for @texnixe to be on board too when she needs to add the next profiles.

texnixe:
As long as we can manage to keep it short, free fields are usually inviting.

texnixe (approved):

PR #2258, opened 2024-03-11
Reduce header margin

distantnative:
Probably a bit controversial, but another suggestion to reduce vertical space a little more, make the site a bit more compact.


bastianallgeier:
I think it works quite well except the buy banner on mobile. But I'm not very happy about that in general tbh.

distantnative:
@bastianallgeier indeed not quite ideal. I added a fix which might work for the time being until we redesign the sales banner?

PR #2260, opened 2024-03-14
Update cookbook-recipe.txt

verkruemelt:
Command `convert` must be added to execute the function properly.


bastianallgeier:
Convert should actually already been added in the convert method. Do you maybe overwrite the bin option in your config? If yes, you need to include the convert command in there

verkruemelt:
No, I did not set the `bin` option at all. I Have set driver to `im`.
But I noticed, if I set the `bin` option expicit to `convert` it works. Maybe because I have to use Windows for now.

Both options semms to be neccessary on Windows to work properly:

```php
return [
	'thumbs' => [
		'autoOrient' => true,
		'driver' => 'im',
		'bin' => 'convert'
]
```

bastianallgeier:
This is not a Windows-specific issue. The convert binary is not always available. It depends on the path settings. That's why we added the bin option. I have to specifically set it on macOS as well.

PR #2261, opened 2024-03-15
Improve `resources` and add to ToC

distantnative:
Better universal handling of `resources` field as "More information" section in articles, incl. support that this section gets included in the table of content.

Added one guide example: Rendering & logic > Templates


bastianallgeier:
I've adjusted the design of the resources a bit to work better with the prev/next links

Screenshot 2024-03-18 at 12 31 15

bastianallgeier (approved):

PR #2276, opened 2024-03-24
Added new name + latest infos about cookie banner plugin

michnhokn:


michnhokn:
Thank you @texnixe. No need to do this on Sunday 😄

PR #2293, opened 2024-04-17
Update cookbook-recipe.txt

verkruemelt:
$this->root() contains the full file name with the original extension.
The extension must be replaced and not concatenated, otherwise the preview file will not be found and will always be regenerated.


texnixe:
I guess the original idea was to keep the complete filename and add the image extension, but then the code to find the file would have to be adapted, and the hook version also cuts off the extension.

verkruemelt:
I was thinking the same thing. I only found the 'error' because I integrated several preview elements on one page and the performance was pretty bad by Kirby standards. 😂

PR #2295, opened 2024-04-17
Add Alphabetise plugin

splorp:


texnixe:
Just trying to understand what this plugin does that cannot be achieved with

```
$groupedElements = $page->children()->listed()->sortBy('title', 'asc')->group(fn ($item) => $item->title()->value()[0]);
```
?

splorp:
Hmmm … after some tinkering on my site, your example does produce the same grouping.

I wasn’t aware of the `groupBy()` and `group()` functions … must have missed that in previous releases.

Were these functions available in K3? If not, then the plugin would still be useful for those installations.

texnixe:
> Were these functions available in K3

Yes, they are available in Kirby 3 as well. We had a whole cookbook recipe about grouping already in Kirby 3: https://v39.getkirby.com/docs/cookbook/content/grouping-collections

There is now an updated version in the Kirby 4 docs, but as you can see, this is really flexible and allows you all sorts of grouping.

I understand this is frustrating after investing work into this, but I wouldn't really want to promote the use of a plugin for something that can be achieved natively.

splorp:
The plugin was originally created for a K2 site by another developer. I had been using it faithfully through K3 and K4 updates and ended up fixing a couple bugs recently. Since the original developer seemed to have dropped support aouple of years ago, perhaps now I know why.

Thanks again for pointing out those functions. (I need to pay better attention, I guess.)

PR #2297, opened 2024-04-20
chore: update kirby caddyfile

pReya:
Added some improvements to my Caddy recipe:
- Add compressed encoding to Caddyfile (really no reason to NOT use this)
- Remove `*.txt` (and `*.md`) from blocked files, because it interfers with `robots.txt` in the base folder by default
- Change handling of blocked paths from `redirect` to actually return a 404 HTTP error and `rewrite` to show Kirby error page (before this, it would fail Kirby's security checks in the Admin panel, because these "blocked" requests were being redirected and still return a `200` code)
- Move `tls` directive into the site block, out of the `common` snippet, because the common snippet is being reused in the `error_handler` (which fails if the `tls` directive is still part of the snippet)

### DIFF
--- a/content/docs/3_cookbook/0_development-deployment/0_caddy/cookbook-recipe.txt
+++ b/content/docs/3_cookbook/0_development-deployment/0_caddy/cookbook-recipe.txt
@@ -133,6 +133,7 @@ import kirby
 root * /usr/share/caddy
 ```
 Here, we're importing/using the two snippets we defined earlier. Also we're setting the root folder for this server, which should point to the absolute filesystem path, where your Kirby files are located.
+<br/>
 
 ```
 tls name@user.com
--- a/content/docs/3_cookbook/0_development-deployment/0_caddy/cookbook-recipe.txt
+++ b/content/docs/3_cookbook/0_development-deployment/0_caddy/cookbook-recipe.txt
@@ -43,19 +43,24 @@ Web server configuration is a very big topic and there are many things to consid
 	php_fastcgi php:9000 # Adjust to your setup
 	tls name@user.com # Adjust to your setup
 	file_server
+	encode zstd gzip
 }
 
 (kirby) {
 	@blocked {
-		path *.txt *.md /content/* /site/* /kirby/* /.*
+		path /content/* /site/* /kirby/* /.*
 	}
-	redir @blocked /
+	error @blocked "Not found" 404
+	handle_errors {
+		rewrite * /error # This should reference your Kirby error page
+		import common
+	}
+	root * /usr/share/caddy # Adjust to your setup
 }
 
 mydomain.com { # Adjust to your setup
 	import common
 	import kirby
-	root * /usr/share/caddy # Adjust to your setup
 }
 
 ```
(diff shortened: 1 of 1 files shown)


content/docs/3_cookbook/0_development-deployment/0_caddy/cookbook-recipe.txt
```diff
@@ -2,7 +2,7 @@ Title: Kirby meets Caddy
 
 ----
 
-Published: 2023-10-28
+Published: 2024-04-20
```
bastianallgeier: @lukasbestle should have a look at the Caddy config changes. I don't know enough about it. But we shouldn't change the published date here.

pReya: Fixed it.

content/docs/3_cookbook/0_development-deployment/0_caddy/cookbook-recipe.txt
```diff
@@ -41,21 +41,26 @@ Web server configuration is a very big topic and there are many things to consid
...
+		path /content/* /site/* /kirby/* /.*
+	}
+	error @blocked "Not found" 404
+	handle_errors {
+		rewrite * /error # This should reference your Kirby error page
+		import common
 	}
-	redir @blocked /
+	root * /usr/share/caddy # Adjust to your setup
+	tls name@user.com # Adjust to your setup
```
lukasbestle: Couldn't we move these two to the site block below? If multiple sites are served by the same Caddy server, these directives might be different for each one.

content/docs/3_cookbook/0_development-deployment/0_caddy/cookbook-recipe.txt
```diff
@@ -77,37 +82,41 @@ We're creating two custom **snippets** here with the names `common` and `kirby`.
...
 <br/>
 
-
 ```
-file_server
+encode zstd gzip
 ```
-
-This directive will make sure Caddy serves all static assets, e.g. all of the JavaScript, font files, icons, images, etc. Without this directive, only PHP files would be served/interpreted.
+We'll enable compression to reduce the transfered file size. This will make your site faster and has been supported by all major browsers for many years.
```
lukasbestle:
```suggestion
We'll enable compression to reduce the transferred file size. This will make your site faster and has been supported by all major browsers for many years.
```

content/docs/3_cookbook/0_development-deployment/0_caddy/cookbook-recipe.txt
```diff
@@ -77,37 +82,41 @@ We're creating two custom **snippets** here with the names `common` and `kirby`.
...
+Here we are doing some Kirby-specific error handling. First, we're defining a custom matcher – which is a statement that we can then use in multiple locations as an argument for other directives. Our matcher is called `@blocked` – because we're defining paths which should not be accessible by website visitors: everything in the `content`, `site` or `kirby` folders, as well as hidden files. In the following line, we're returning a 404 HTTP error, when any of these files are being requested.
 <br/>
 
+```
+handle_errors {
+	rewrite * /error # This should reference your Kirby error page
+	import common
+}
+```
+The error we're returning from the line before will result in the browser showing an empty page – which is typically not what we want. Instead, we want to show Kirbys error page, which is served under the `/error` path by default. So we're rewriting all the requests coming into this error handler to the Kirby error page. And we also need the `common` block here, so PHP will be executed as well (because Kirbys error page is not a static HTML file, but is dynamically generated via PHP).
```
lukasbestle:
```suggestion
The error we're returning from the line before will result in the browser showing an empty page – which is typically not what we want. Instead, we want to show Kirby's error page, which is served under the `/error` path by default. So we're rewriting all the requests coming into this error handler to the Kirby error page. And we also need the `common` block here, so PHP will be executed as well (because Kirby's error page is not a static HTML file, but is dynamically generated via PHP).
```

content/docs/3_cookbook/0_development-deployment/0_caddy/cookbook-recipe.txt
```diff
@@ -123,5 +132,10 @@ import common
 import kirby
 root * /usr/share/caddy
 ```
-
 Here, we're importing/using the two snippets we defined earlier. Also we're setting the root folder for this server, which should point to the absolute filesystem path, where your Kirby files are located.
+
```
lukasbestle, line 137:
```suggestion
<br/>

```

lukasbestle (commented):
Sorry for the delay in review and thank you very much for your great work on these recipes. 💛

lukasbestle:
Thank you very much 👍

PR #2298, opened 2024-04-22
Redirects to new plugins directory

bastianallgeier:
A few things to note: 

- I put all redirects into a separate route file to keep this a bit more maintainable. 
- I wasn't able to remove the plugins content folder yet because we still have references in some places where we include them. E.g. partners list their plugins and the 2021 page also still has a list of plugins, which would break. I think the 2021 one page can be hard-coded or we could remove the plugins. It's historical anyway. But for the partner profiles we need to find a good way to pull the data from the directory somehow.


lukasbestle (approved):
Made some additional changes. Please take another look before merging.

Additional todos:
- Consider replacing the Matomo plugin on "Features for Developers" as that's only for v3.
- Update the update check in Kirby to request the plugin JSONs from the plugins subdomain directly. We may need KeyCDN there to avoid a lot of requests to our server. Maybe we even need to do this before the launch, not sure.

PR #2300, opened 2024-05-01
Update reference-classmethod.txt of Page::create()

elsonigo:
Added "draft" to $prop to example.


texnixe:
@elsonigo Thanks for your PR. However, instead of changing what the example does, it would IMO make more sense to add all possible props, because there are more.

PR #2316, opened 2024-05-26
Docs on how to modify Sane classes

### LINKED ISSUE #1573: Add information about overriding Sane class defaults
This is totally niche, and probably very low priority – just opening this issue for future consideration, as I've run into this twice now.

The `Sane` class comes with a preset of permitted SVG/XML attributes etc., which can be overridden. These allowlist overrides are not currently documented (and, frankly, I'm not quite sure where that info would be best placed in the reference).

Related:
- https://forum.getkirby.com/t/how-to-define-my-allowlist-for-svg-attributes-k3-6-sane-class/23999/6
- https://github.com/getkirby/kirby/issues/3433#issuecomment-864851723

distantnative:


lukasbestle (approved):
Thank you 👍

PR #2318, opened 2024-05-31
Move to CDN

bastianallgeier:


site/snippets/layouts/head.php
```diff
@@ -13,51 +13,15 @@
...
 <?php if($page->id() === 'buy'): ?>
 <link rel="preload" href="<?= url('buy/prices') ?>" as="fetch" />
 <?php endif ?>
 
-<?php if($page->id() === 'meet'): ?>
-<link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css" integrity="sha256-p4NxAoJBhIIN+hmNHrzRCf9tD/miZyoHS5obTRR9BMY="      crossorigin=""/>
-<link rel="stylesheet" href="https://unpkg.com/leaflet.markercluster@1.4.1/dist/MarkerCluster.css"/>
-<script src="https://unpkg.com/leaflet@1.9.4/dist/leaflet.js" integrity="sha256-20nQCchB9co0qIjJZRGuk2/Z9VM+kNiyxNV1lvTlZBo=" crossorigin=""></script>
-<script src="https://unpkg.com/leaflet.markercluster@1.4.1/dist/leaflet.markercluster.js"></script>
-<?php endif ?>
```
bastianallgeier, line 58: Oh, I didn't even realize we still had this in here. Good riddance!

bastianallgeier (commented):
I really appreciate the accessibility fixes, but it's also making the CDN parts harder to review. 

As a learning from the panel, I think it was a mistake to adopt specific color variable names and utilities in the past. We should use --color-text, --color-text-dimmed and other color value independent variables and utilities more. Otherwise it is getting really hard to change the color values without also breaking the variable names. 

The thing here is that you basically need an additional utility and variable name for rather dark but not entirely black text on our gray background. I wasn't able to come up with a good variable name for it so far. 

I wonder if we should still keep your changes for this PR and then maybe find something better afterwards. 

The rest of the PR is really nice!

lukasbestle:
Those last commits came from a systematic Lighthouse check of our templates. Typical “while I was at it”. I included those in the `develop` branch so I could verify the results with the CDN. But of course not suitable for a PR review.

I can take that contrast commit out of the branch and create a separate PR. Makes it easier to review because we will see all changed places with those gray boxes.

About the variable name: Maybe `--color-text-dimmed-dark`? It’s quite long but IMO at least obvious about its purpose next to the existing variables.

bastianallgeier:
I think it's fine. Let's move more utility changes to another PR. It's good that the a11y issues are fixed by this and it's no big deal to adjust it some more later.

bastianallgeier:
Ironically, now you have to approve your own PR, because I've created it 🤣

lukasbestle:
Will merge it manually

PR #2320, opened 2024-06-01
Use final URLs after redirects in menus

lukasbestle:
Some of our templates redirect to another page (e.g. child page). If we include those pages in menus, the user needs to be redirected even though we already know the final target. This is currently mostly the case with `docs/guide` and `license`.

This PR implements two new page methods:

- `menuUrl()`: Returns the final URL after redirects to be used in menus. Overridden in the page model where relevant, otherwise falls back to the page URL.
- `isActiveInMenu()`: Variant of `isActive()` that respects `menuUrl()`. This fixes a bug in the menu for legal pages where the [License page](https://getkirby.com/license) wouldn't show an active state because it immediately redirects to its last child.

### DIFF
--- a/site/models/buzz-entry.php
+++ b/site/models/buzz-entry.php
@@ -1,10 +1,11 @@
 <?php
 
-use Kirby\Cms\Page;
 use Kirby\Content\Field;
 use Kirby\Toolkit\Str;
 
-class BuzzEntryPage extends Page
+require_once __DIR__ . '/default.php';
+
+class BuzzEntryPage extends DefaultPage
 {
 	public function blurb(): Field
 	{
--- a/site/models/buzz-entry.php
+++ b/site/models/buzz-entry.php
@@ -19,11 +19,6 @@ class BuzzEntryPage extends DefaultPage
 			Str::startsWith($url, $this->site()->url()) === false;
 	}
 
-	public function menuUrl(): string
-	{
-		return $this->url();
-	}
-
 	public function url($options = null): string
 	{
 		if ($this->link()->isNotEmpty() === true) {
--- a/site/models/buzz-entry.php
+++ b/site/models/buzz-entry.php
@@ -18,6 +18,11 @@ class BuzzEntryPage extends Page
 			Str::startsWith($url, $this->site()->url()) === false;
 	}
 
+	public function menuUrl(): string
+	{
+		return $this->url();
+	}
+
 	public function url($options = null): string
 	{
 		if ($this->link()->isNotEmpty() === true) {
--- a/site/models/buzz-entry.php
+++ b/site/models/buzz-entry.php
@@ -3,8 +3,6 @@
 use Kirby\Content\Field;
 use Kirby\Toolkit\Str;
 
-require_once __DIR__ . '/default.php';
-
 class BuzzEntryPage extends DefaultPage
 {
 	public function blurb(): Field
--- a/site/models/glossary-entry.php
+++ b/site/models/glossary-entry.php
@@ -6,6 +6,6 @@ class GlossaryEntryPage extends DefaultPage
 {
 	public function menuUrl(): string
 	{
-		return $this->link()->or($this->url())->value();
+		return $this->link()->toUrl() ?? $this->url();
 	}
 }
--- a/site/models/glossary-entry.php
+++ b/site/models/glossary-entry.php
@@ -1,7 +1,5 @@
 <?php
 
-require_once __DIR__ . '/default.php';
-
 class GlossaryEntryPage extends DefaultPage
 {
 	public function menuUrl(): string
--- a/site/models/glossary-entry.php
+++ b/site/models/glossary-entry.php
@@ -2,15 +2,10 @@
 
 use Kirby\Cms\Page;
 
-class ReleaseGuidePage extends Page
+class GlossaryEntryPage extends Page
 {
 	public function menuUrl(): string
 	{
 		return $this->link()->or($this->url())->value();
-	}
-
-	public function release()
-	{
-		return $this->parents()->not('releases')->last();
 	}
 }
--- a/site/models/glossary-entry.php
+++ b/site/models/glossary-entry.php
@@ -1,8 +1,8 @@
 <?php
 
-use Kirby\Cms\Page;
+require_once __DIR__ . '/default.php';
 
-class GlossaryEntryPage extends Page
+class GlossaryEntryPage extends DefaultPage
 {
 	public function menuUrl(): string
 	{
--- a/site/models/link.php
+++ b/site/models/link.php
@@ -6,6 +6,6 @@ class LinkPage extends DefaultPage
 {
 	public function menuUrl(): string
 	{
-		return $this->link()->value();
+		return url($this->link()->value());
 	}
 }
--- a/site/models/link.php
+++ b/site/models/link.php
@@ -2,15 +2,10 @@
 
 use Kirby\Cms\Page;
 
-class ThemePage extends Page
+class LinkPage extends Page
 {
 	public function menuUrl(): string
 	{
 		return $this->link()->value();
-	}
-
-	public function prio()
-	{
-		return parent::prio()->or(9999);
 	}
 }
--- a/site/models/link.php
+++ b/site/models/link.php
@@ -1,7 +1,5 @@
 <?php
 
-require_once __DIR__ . '/default.php';
-
 class LinkPage extends DefaultPage
 {
 	public function menuUrl(): string
--- a/site/models/link.php
+++ b/site/models/link.php
@@ -6,6 +6,6 @@ class LinkPage extends DefaultPage
 {
 	public function menuUrl(): string
 	{
-		return url($this->link()->value());
+		return $this->link()->toUrl();
 	}
 }
--- a/site/models/link.php
+++ b/site/models/link.php
@@ -1,8 +1,8 @@
 <?php
 
-use Kirby\Cms\Page;
+require_once __DIR__ . '/default.php';
 
-class LinkPage extends Page
+class LinkPage extends DefaultPage
 {
 	public function menuUrl(): string
 	{
--- a/site/models/reference-quicklink.php
+++ b/site/models/reference-quicklink.php
@@ -49,14 +49,6 @@ class ReferenceQuickLinkPage extends DefaultPage
 		return $children;
 	}
 
-	/**
-	 * Returns whether the referenced page is open
-	 */
-	public function isOpen(): bool
-	{
-		return $this->link()->toPage()?->isOpen() ?? false;
-	}
-
 	public function template(): Template
 	{
 		return $this->kirby()->template('link');
--- a/site/models/reference-quicklink.php
+++ b/site/models/reference-quicklink.php
@@ -49,6 +49,11 @@ class ReferenceQuickLinkPage extends DefaultPage
 		return $children;
 	}
 
+	public function menuUrl(): string
+	{
+		return url($this->link()->value());
+	}
+
 	public function template(): Template
 	{
 		return $this->kirby()->template('link');
--- a/site/models/reference-quicklink.php
+++ b/site/models/reference-quicklink.php
@@ -1,12 +1,13 @@
 <?php
 
-use Kirby\Cms\Page;
 use Kirby\Cms\Pages;
 use Kirby\Content\Field;
 use Kirby\Template\Template;
 use Kirby\Toolkit\Str;
 
-class ReferenceQuickLinkPage extends Page
+require_once __DIR__ . '/default.php';
+
+class ReferenceQuickLinkPage extends DefaultPage
 {
 	/**
 	 * Creates children collection from `menu` content field
--- a/site/models/reference-quicklink.php
+++ b/site/models/reference-quicklink.php
@@ -5,8 +5,6 @@ use Kirby\Content\Field;
 use Kirby\Template\Template;
 use Kirby\Toolkit\Str;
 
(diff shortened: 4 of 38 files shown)


site/plugins/site/extensions/pageMethods.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+return [
+	/**
+	 * Checks whether the page is the current page in a menu context,
+	 * considering redirects to other pages
+	 */
+	'isActiveInMenu' => function (bool $hasSubmenu = false) {
```
distantnative, line 8: Wondering if `->isOpen()` wouldn't work just as well? https://getkirby.com/docs/reference/objects/cms/page/is-open

distantnative:
@lukasbestle Could you please share why we need to replace `->isOpen()` with `->isActiveInMenu()`? Based on the description above, I would this that `isOpen` is exactly made for that.

And how about if we overwrite the `url()` method for those instead of creating a new `menuUrl()` for all object? Do you see a case where those models still need their original `url()`?

lukasbestle:
@distantnative On `isOpen()`: That's basically the opposite of what we need. We explicitly don't want to highlight the main menu item if a subpage is active. We do use `isOpen` for controlling whether the menu is expanded by default, but the other use we had of `isOpen` for the submenu was wrong IMO. Inside a submenu, we are already displaying the page itself and there are no additional children. Or have I misunderstood the structure?

About replacing the `url()` method: That's what I had before, but unfortunately it won't work for this use case. If the parent returns the URL of its child, the child's `url()` method then causes an infinite loop because it uses its parent's `url()` as a base for adding its own UID at the end.

lukasbestle:
Thinking about it again, I think you are right. We don't have recursive submenus, only two levels (main and one sub level). So in the sub level, the menu item should be highlighted even if one of its children is active. So `isOpen()` is correct there. And I can't think of a case when that would be broken by redirects.

site/plugins/site/extensions/pageMethods.php
```diff
@@ -0,0 +1,26 @@
...
+		// otherwise use the menu URL for the comparison
+		return $this->menuUrl() === $this->site()->page()?->url();
+	},
+
+	/**
+	 * Final URL after redirects to be used in menus;
+	 * fallback if the page model doesn't override it
+	 */
+	'menuUrl' => fn () => $this->url()
+];
```
distantnative: `Suggestion:` We could use a default page model instead (https://getkirby.com/docs/guide/templates/page-models#default-page-model) - that way those would be together with the other page models in `site/models` which could make it easier to understand/keep an overview later down the road.

Would though include that the other page models have to extend the `DefaultPage` model then.

site/models/buzz-entry.php
```diff
@@ -18,6 +18,11 @@ public function isExternalLink(): bool
 			Str::startsWith($url, $this->site()->url()) === false;
 	}
 
+	public function menuUrl(): string
+	{
+		return $this->url();
+	}
+
```
distantnative:
```suggestion
```

`Todo:` This looks to be the default as in the custom page method (potential default page model). So I don't think we need to add this here.

site/models/terms.php
```diff
@@ -0,0 +1,24 @@
...
+		if ($latest?->intendedTemplate()->name() === 'terms') {
+			return $latest;
+		}
+
+		return null;
+	}
+
+	public function menuUrl(): string
+	{
+		return $this->latestVersion()->url() ?? $this->url();
```
distantnative:
```suggestion
		return $this->latestVersion()?->url() ?? $this->url();
```

site/snippets/layouts/footer.php
```diff
@@ -80,11 +80,11 @@
...
-							<li><a href="/security">Security</a></li>
-							<li><a href="/privacy">Privacy</a></li>
-							<li><a href="/license">License</a></li>
-							<li><a href="/press">Presskit</a></li>
-							<li><a href="/contact">Contact</a></li>
+							<li><a href="<?= page('security')->menuUrl() ?>">Security</a></li>
+							<li><a href="<?= page('privacy')->menuUrl() ?>">Privacy</a></li>
+							<li><a href="<?= page('license')->menuUrl() ?>">License</a></li>
+							<li><a href="<?= page('press')->menuUrl() ?>">Presskit</a></li>
+							<li><a href="<?= page('contact')->menuUrl() ?>">Contact</a></li>
```
distantnative: `Suggestion:` To keep the footer and main menus DRY-er, we could use a snippet which we just pass something like

```php
[
    'For developers' => page('features/developers'),
    'For designers' => page('features/designers'),
    'Feedback' => 'https://feedback.getkirby.com'
]
```

And the snippet renders each `<li>` depending whether it contains a page object or a string (for external) differently).

site/models/buzz-entry.php
```diff
@@ -1,10 +1,11 @@
 <?php
 
-use Kirby\Cms\Page;
 use Kirby\Content\Field;
 use Kirby\Toolkit\Str;
 
-class BuzzEntryPage extends Page
+require_once __DIR__ . '/default.php';
```
distantnative: `Suggestion:` Instead of having to add it to each page model, we can add this line in the `config.php` or `site` plugin's `index.php`.

site/models/glossary-entry.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+require_once __DIR__ . '/default.php';
+
+class GlossaryEntryPage extends DefaultPage
+{
+	public function menuUrl(): string
+	{
+		return $this->link()->or($this->url())->value();
```
distantnative:
```suggestion
		return $this->link()->toUrl() ?? $this->url();
```

For absolute URLs. And using `->toUrl()`.

site/models/link.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+require_once __DIR__ . '/default.php';
+
+class LinkPage extends DefaultPage
+{
+	public function menuUrl(): string
+	{
+		return url($this->link()->value());
```
distantnative:
```suggestion
		return $this->link()->toUrl();
```

site/models/reference-quicklink.php
```diff
@@ -48,12 +49,9 @@ public static function childrenFromContentField(Field $field): array
...
 	}
 
-	/**
-	 * Returns whether the referenced page is open
-	 */
-	public function isOpen(): bool
+	public function menuUrl(): string
 	{
-		return $this->link()->toPage()?->isOpen() ?? false;
+		return url($this->link()->value());
```
distantnative:
```suggestion
		return $this->link()->toUrl();
```

site/models/release-guide.php
```diff
@@ -1,9 +1,14 @@
...
 
-use Kirby\Cms\Page;
+require_once __DIR__ . '/default.php';
 
-class ReleaseGuidePage extends Page
+class ReleaseGuidePage extends DefaultPage
 {
+	public function menuUrl(): string
+	{
+		return $this->link()->or($this->url())->value();
```
distantnative:
```suggestion
		return $this->link()->toUrl() ?? $this->url();
```

site/snippets/layouts/menu-items.php
```diff
@@ -0,0 +1,9 @@
+<?php foreach ($items as $label => $target): ?>
+	<?php if ($target instanceof \Kirby\Cms\Page): ?>
+	<li><a href="<?= $target->menuUrl() ?>"><?= $label ?></a></li>
+	<?php elseif (is_string($label) === true): ?>
+	<li><a <?php if ($externalClass ?? null): ?>class="<?= $externalClass ?>" <?php endif ?>href="<?= $target ?>"><?= $label ?></a></li>
```
distantnative:
```suggestion
	<li><a <?= e($externalClass ?? null, 'class="' . $externalClass . '" ') ?>href="<?= $target ?>"><?= $label ?></a></li>
```

site/snippets/layouts/menu.php
```diff
@@ -6,67 +6,86 @@
...
-					<li><a href="/releases">Releases</a></li>
-					<li><a class="is-external" href="https://feedback.getkirby.com">Feedback</a></li>
+					<?php snippet('layouts/menu-items', [
+						'externalClass' => 'is-external',
+						'items' => [
+							'For developers' => page('features/developers'),
+							'For designers' => page('features/designers'),
+							'For content creators' => page('features/creators'),
+							'For clients & agencies' => page('features/clients'),
+							'hr',
```
distantnative:
```suggestion
							'-',
```

Would be more consistent with Kirby itself (e.g. Panel menu).

site/snippets/layouts/menu.php
```diff
@@ -6,67 +6,86 @@
...
-					<li><a href="/kosmos">Newsletter</a></li>
-					<li><a href="/buzz">Buzz</a></li>
-					<li><hr /></li>
-					<li><a class="is-external" href="https://hub.getkirby.com">License Hub</a></li>
+					<?php snippet('layouts/menu-items', [
+						'externalClass' => 'is-external',
+						'items' => [
+							'Plugins' => 'https://plugins.getkirby.com',
+							'Themes' => page('themes'),
+							'hr',
```
distantnative:
```suggestion
							'-',
```

site/snippets/layouts/menu.php
```diff
@@ -6,67 +6,86 @@
...
+							'Newsletter' => page('kosmos'),
+							'Buzz' => page('buzz'),
+							'hr',
```
distantnative:
```suggestion
							'-',
```

site/snippets/layouts/menu.php
```diff
@@ -6,67 +6,86 @@
...
-					<li><a class="is-external" href="https://community.getkirby.com">Community map</a></li>
-					<li><hr /></li>
-					<li><a class="is-external" href="https://mastodon.social/@getkirby">Mastodon</a></li>
-					<li><a class="is-external" href="https://www.linkedin.com/company/getkirby">LinkedIn</a></li>
-					<li><a class="is-external" href="https://instagram.com/getkirby">Instagram</a></li>
+					<?php snippet('layouts/menu-items', [
+						'externalClass' => 'is-external',
+						'items' => [
+							'Get together' => page('meet'),
+							'hr',
```
distantnative:
```suggestion
							'-',
```

site/snippets/layouts/menu.php
```diff
@@ -6,67 +6,86 @@
...
+							'Support forum' => 'https://forum.getkirby.com',
+							'Discord chat' => 'https://chat.getkirby.com',
+							'Community map' => 'https://community.getkirby.com',
+							'hr',
```
distantnative:
```suggestion
							'-',
```

site/snippets/layouts/menu-items.php
```diff
@@ -0,0 +1,9 @@
+<?php foreach ($items as $label => $target): ?>
+	<?php if ($target instanceof \Kirby\Cms\Page): ?>
+	<li><a href="<?= $target->menuUrl() ?>"><?= $label ?></a></li>
+	<?php elseif (is_string($label) === true): ?>
+	<li><a <?= e($externalClass ?? null, 'class="' . $externalClass . '" ') ?>href="<?= $target ?>"><?= $label ?></a></li>
```
distantnative: `Issue:` `$externalClass` is never passed where the snippet is used. I would actually suggest to drop that variable and always add the class `is-external` here.

```suggestion
	<li><a class="is-external" href="<?= $target ?>"><?= $label ?></a></li>
```

site/models/reference-quicklink.php
```diff
@@ -48,12 +47,9 @@ public static function childrenFromContentField(Field $field): array
 		return $children;
 	}
 
-	/**
-	 * Returns whether the referenced page is open
-	 */
-	public function isOpen(): bool
```
distantnative, line 54: `Issue:` Something seems to have been lost here. E.g. https://getkirby.test/docs/reference/objects/filesystem/dir/is-readable - on `main` "Dir" in the "Tools" section in the Reference sidebar is highlighted. With this PR it isn't anymore. Since `->isOpen()` is still used in some parts of the sidebar, we probably cannot remove this overwritten method here.

distantnative (changes requested):
Code looks great, but ran into 2 errors.

site/snippets/layouts/menu-items.php
lukasbestle: It is (in the main menu). We cannot hardcode the class as the footer menu shouldn't have it.

The issue was caused by the use of `e()` that makes the string in the second argument already evaluate even if the variable is empty or doesn't exist (unlike the `if` we had before). Fixed it with another fallback. But TBH, by that point I would prefer to go back to the original version with the `if`.

distantnative: Why shouldn't the footer menu have it?

distantnative:
"Changes requested" shows up so harsh and negative. I'll keep using "Comments" again going forward. Just wanted to indicate that those issues need to be solved before we can merge, but the UI makes it look so negative.

site/snippets/layouts/menu-items.php
lukasbestle: It probably doesn't harm, but we don't use the class for anything there. In the main menu it controls the arrow icon, the footer doesn't have that.

distantnative: Yes, that's why I think it would be beneficial to just include it for the footer as well. It simplifies the logic in the `menu-item` snippet quite nicely and doesn't affect (the look of) the footer currently. But if we ever would want to differentiate those, this class would also make sense in the context of the footer.

distantnative (approved):

PR #2322, opened 2024-06-01
Home: improve delay when switching story images

distantnative:
- Decreased the CSS transition time
- Removed the initial JS delay
- Shortened the final JS delay to match CSS transition time

### DIFF
--- a/assets/css/layouts/home.css
+++ b/assets/css/layouts/home.css
@@ -95,10 +95,11 @@
 	transition: all 0.2s;
 }
 .playground-header-figure-wrapper img {
-	transition: all 0.1s;
+	transition: opacity 0.4s;
 }
 .playground-header-figure-wrapper.loading img {
 	opacity: 0;
+	transition: opacity 0.2s;
 }
 
 /** Mobile */
--- a/assets/css/layouts/home.css
+++ b/assets/css/layouts/home.css
@@ -95,7 +95,7 @@
 	transition: all 0.2s;
 }
 .playground-header-figure-wrapper img {
-	transition: all 0.3s;
+	transition: all 0.1s;
 }
 .playground-header-figure-wrapper.loading img {
 	opacity: 0;
--- a/assets/css/layouts/home.css
+++ b/assets/css/layouts/home.css
@@ -92,9 +92,9 @@
 	display: block;
 	background: var(--color-light);
 	overflow: hidden;
-	transition: all 0.2s;
 }
 .playground-header-figure-wrapper img {
+	opacity: 1;
 	transition: opacity 0.4s;
 }
 .playground-header-figure-wrapper.loading img {
--- a/assets/js/layouts/playground.js
+++ b/assets/js/layouts/playground.js
@@ -12,30 +12,40 @@ export class Playground {
 		});
 	}
 
+	async loadHtml(link) {
+		const response = await fetch(link);
+		const body = await response.text();
+		const parser = new DOMParser();
+		const doc = parser.parseFromString(body, "text/html");
+
+		// load the new image as hidden before it gets injected
+		doc
+			.querySelector(".playground-header-figure-wrapper")
+			.classList.add("loading");
+
+		return doc;
+	}
+
 	async switchTo(link) {
 		// fade out the old image
 		this.$el
 			.querySelector(".playground-header-figure-wrapper")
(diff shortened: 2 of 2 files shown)


assets/js/layouts/playground.js
```diff
@@ -11,34 +9,37 @@ export class Playground {
...
+			.classList.add("loading");
+
+		// replace the playground
+		this.$el.innerHTML = doc.querySelector(".playground").innerHTML;
+
+		// fade in the image on load
+		this.$el
+			.querySelector(".playground-header-figure img")
+			.addEventListener("load", async function () {
+				await new Promise((resolve) => setTimeout(resolve, 100));
```
lukasbestle: TBH I don't fully understand what this delay is even trying to do, also in the old version. We had one delay that waits until the old image is faded out and one delay that waits for another 200 (now 100) ms before fading the new image in. But the actual fading-in happens *after* that second delay. So why do we need to wait here?

What I think we need is logic like "perform the fetch immediately and if less than 100 ms (= fading-out time of the old image) has passed by the time we are ready, wait until the fade-out was completed". So basically a `Promise.all()` of the fetch request and a 100 ms delay. Once both are finished, the new content can be injected and faded in.

assets/css/layouts/home.css
```diff
@@ -95,7 +95,7 @@
 	transition: all 0.2s;
 }
 .playground-header-figure-wrapper img {
-	transition: all 0.3s;
+	transition: all 0.1s;
 }
 .playground-header-figure-wrapper.loading img {
 	opacity: 0;
```
lukasbestle, line 101: Something is wrong about the transition. If I simulate a slow connection (200 ms extra wait time for the `fetch` request), I get flickering like this in Safari:

https://github.com/getkirby/getkirby.com/assets/1595007/82687b9d-70b9-4a29-afc4-ef0ce7ab8174

Other browsers don't seem to have that issue, but in Safari it happens all the time. Do you also see that?

lukasbestle:
About preloading: I don't think it makes sense to preload all playground stories when visiting the homepage, but we could make it a bit faster by preloading the main story image immediately in the `switchTo` method when a new story is loaded.

Could work like this:

1. Add a custom `data-image` attribute to the [link in the menu](https://github.com/getkirby/getkirby.com/blob/de244bc76db0c2b1e08966d843b025ee003f8423/site/snippets/templates/home/playground/header.php#L38).
2. Before starting the `fetch` request for the HTML, do a `new Image().src = url;` to preload the image in the background.

assets/css/layouts/home.css
```diff
@@ -95,7 +95,7 @@
 	transition: all 0.2s;
 }
 .playground-header-figure-wrapper img {
-	transition: all 0.3s;
+	transition: all 0.1s;
```
lukasbestle: Testing this some more, I think I prefer 200 ms for the transition. It feels a bit more calm and also leaves us more time to load the HTML and main image.

assets/js/layouts/playground.js
distantnative: You are right, and the `Promise.all()` sounds like a good direction. But I think it's not just fetch and the animation delay, but the image being loaded additionally.

distantnative: Ok, now fully understood it. Give me a few minutes.

distantnative:
@lukasbestle I am also thinking if there would be a better way than fetch the whole home page when switching, just to extract the playground part. Maybe we could work with a content representation here that only return the playground part?

assets/css/layouts/home.css
distantnative: I have it in Firefox as well that sometime the fade-in animation is shown, but sometimes not at all.

lukasbestle: Splitting the image loading and the fade-in to two separate ticks seems to do the trick.

lukasbestle:
Made some additional enhancements to reduce flickering. I think we could merge it like this already, check the results in production and optimize further if needed. I feel like reducing the HTML payload would bring us a tiny improvement compared to the added complexity as it's just text and normal pages load quickly.

distantnative:
I still experience sometimes the flicker. But it's the same on the old version. So still purely improvement. I agree on merging already and tweaking further later.

PR #2325, opened 2024-06-02
Additional performance improvements for home playground

lukasbestle:
## What was changed

- Move the styles for the main image one layer up so that the `img` itself doesn't need to be transitioned
- Introduce a slight delay before fading in the image after the DOM change and the image load event
- Give the frontend the parts of the media URL that change for each image so it can dynamically build the correct image URL (respecting the `srcset` variant the browser selected for the current image)

## Why was it changed

- Reduce flickering in Safari and Firefox, respectively (at least I can now reproduce the flickering issues much less often, even when simulating network delays)
- Reduce loading delay when the user switches to another playground item as the image can be loaded already while waiting for and parsing the HTML

### DIFF
--- a/assets/js/layouts/playground.js
+++ b/assets/js/layouts/playground.js
@@ -28,8 +28,7 @@ export class Playground {
 
 	async switchTo(link) {
 		// fade out the old image
-		this.$el
-			.querySelector(".playground-header-figure-wrapper")
+		this.currentWrapper()
 			.classList.add("loading");
 
 		// since our CSS transition to fade out the image takes 200ms,
@@ -43,8 +42,7 @@ export class Playground {
 		this.$el.innerHTML = doc.querySelector(".playground").innerHTML;
 
 		// fade in the image once loaded
-		this.$el
-			.querySelector(".playground-header-figure img")
+		this.currentImg()
 			.addEventListener("load", function () {
 				// let the browser render the image first to reduce flickering issues
 				setTimeout(() => this.parentNode.classList.remove("loading"), 10);
@@ -52,5 +50,13 @@ export class Playground {
 
 		// reactivate code highlighting
 		Prism.highlightAll();
+	}
+
+	currentWrapper() {
+		return this.$el.querySelector(".playground-header-figure-wrapper");
+	}
+
+	currentImg() {
+		return this.$el.querySelector(".playground-header-figure-wrapper img");
 	}
 }
--- a/assets/js/layouts/playground.js
+++ b/assets/js/layouts/playground.js
@@ -46,8 +46,8 @@ export class Playground {
 		this.$el
 			.querySelector(".playground-header-figure img")
 			.addEventListener("load", function () {
-				// use next tick to avoid flickering issues
-				setTimeout(() => this.parentNode.classList.remove("loading"), 0);
+				// let the browser render the image first to reduce flickering issues
+				setTimeout(() => this.parentNode.classList.remove("loading"), 10);
 			});
 
 		// reactivate code highlighting
--- a/assets/js/layouts/playground.js
+++ b/assets/js/layouts/playground.js
@@ -12,6 +12,10 @@ export class Playground {
 		});
 	}
 
+	get image() {
+		return this.$el.querySelector(".playground-header-figure-wrapper img");
+	}
+
 	async loadHtml(link) {
 		const response = await fetch(link);
 		const body = await response.text();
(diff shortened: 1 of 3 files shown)


assets/js/layouts/playground.js
```diff
@@ -26,12 +26,17 @@ export class Playground {
...
-		this.$el
-			.querySelector(".playground-header-figure-wrapper")
+		this.currentWrapper()
 			.classList.add("loading");
 
+		// preload the new image
+		const imgSource = this.currentImg()
+			.currentSrc
+			.replace(this.currentWrapper().dataset.image, target.dataset.image);
+		new Image().src = imgSource;
```
distantnative: `Suggestion:`

```suggestion
		new Image().src = this.image.currentSrc.replace(
			this.wrapper.dataset.image, 
			target.dataset.image
		);
```

Depends on suggested change to getters, see below.

`Suggestion:` Could be moved to a separate method `preload(image)`.

assets/js/layouts/playground.js
```diff
@@ -43,14 +48,21 @@ export class Playground {
...
+				setTimeout(() => this.parentNode.classList.remove("loading"), 10);
 			});
 
 		// reactivate code highlighting
 		Prism.highlightAll();
 	}
+
+	currentWrapper() {
+		return this.$el.querySelector(".playground-header-figure-wrapper");
+	}
```
distantnative: `Suggestion:`

```suggestion
	get wrapper() {
		return this.$el.querySelector(".playground-header-figure-wrapper");
	}
```

assets/js/layouts/playground.js
```diff
@@ -43,14 +48,21 @@ export class Playground {
...
+
+	currentImg() {
+		return this.$el.querySelector(".playground-header-figure-wrapper img");
+	}
```
distantnative: `Suggestion:`

```suggestion
	get image() {
		return this.$el.querySelector(".playground-header-figure-wrapper img");
	}
```

assets/js/layouts/playground.js
```diff
@@ -43,14 +48,21 @@ export class Playground {
 		this.$el.innerHTML = doc.querySelector(".playground").innerHTML;
 
 		// fade in the image once loaded
-		this.$el
-			.querySelector(".playground-header-figure img")
+		this.currentImg()
 			.addEventListener("load", function () {
```
distantnative: `Suggestion:`

```suggestion
		this.image.addEventListener("load", function () {
```

Depends on suggested change to getters, see below.

assets/js/layouts/playground.js
```diff
@@ -26,12 +26,17 @@ export class Playground {
 		return doc;
 	}
 
-	async switchTo(link) {
+	async switchTo(link, target) {
 		// fade out the old image
-		this.$el
-			.querySelector(".playground-header-figure-wrapper")
+		this.currentWrapper()
 			.classList.add("loading");
```
distantnative: `Suggestion:`

```suggestion
		this.wrapper.classList.add("loading");
```

distantnative (commented):
Looks good! Added some suggestions for the JS code. Ideally, we would also sort the method/getters alphabetically.

`Thought:` Wondering if the loading change could be less drastic if the loading background color would not be so bright/light (and thus less different to the images).

assets/js/layouts/playground.js
```diff
@@ -26,11 +30,21 @@ export class Playground {
...
 	}
 
-	async switchTo(link) {
+	preload(target) {
+		const newUrl = this.image.currentSrc.replace(
+			this.wrapper.dataset.image,
+			target.dataset.image
+		);
+
+		new Image().src = newUrl;
```
distantnative, line 39: `Nitpick:` Is there a reason why you want to first assign it to a variable and not directly to the Image?

```suggestion
		new Image().src = this.image.currentSrc.replace(
			this.wrapper.dataset.image,
			target.dataset.image
		);
```

distantnative (approved):
Just one small question. I'll approve it already as either way is fine.

Not sure if the dark background is better after all. Torn on that one.

assets/js/layouts/playground.js
lukasbestle: I think it's clearer because when reading the code, you can separate the lines into "one builds the URL and the other preloads". If both are combined, it's harder to follow what's happening, especially because the URL logic is already complex.

lukasbestle:
Yeah, coming back to it, I'm also no longer sure about the background color. The light color has the advantage that it's already in the background of the Panel screenshots, so overall less drastic I suppose. Let's keep it like that for now, we can still change it.

PR #2326, opened 2024-06-02
Events: localize datetime for user's timezone

### LINKED ISSUE #2324: Browser timezone for meetups
The [Get together](https://getkirby.com/meet) page lists upcoming meetups. The time is hardcoded to be formatted in CET.

**Issue:** Since it is summer time at the moment, that could be confusing as people might misinterpret the time as the actual event time, making the time one hour off.

**Possible solution:** We could use UTC as the default time and then progressively enhance on the frontend by replacing the date and time with the value in the visitor's timezone.

distantnative:
### This PR changes...
- Event datetime is stored  in content as raw format retrieved from Discord API
- Adds custom HTML element `<localized-datetime>` which converts its UTC content to a localized string based on the user's timezone

Screenshot 2024-06-02 at 19 56 04

### Reason
Replaces the hardcoded `CET` timezone that currently would display times by 1 hour offset (as we are in CEST now).
- https://github.com/getkirby/getkirby.com/issues/2324

### DIFF
--- a/site/models/meet-events.php
+++ b/site/models/meet-events.php
@@ -18,7 +18,6 @@ class MeetEventsPage extends Page
 		$events   = $this->getData();
 
 		foreach ($events as $event) {
-			$timezone   = 'CET';
 			$children[] = [
 				'slug'     => Str::kebab($event['name']),
 				'num'      => 0,
@@ -28,8 +27,7 @@ class MeetEventsPage extends Page
 				'content'  => [
 					'title'    => $event['name'],
 					'link'     => $event['entity_metadata']['location'],
-					'date'     => Date::optional($event['scheduled_start_time'])->setTimezone(new DateTimeZone($timezone))->format('Y-m-d H:i:s'),
-					'timezone' => $timezone
+					'date'     => Date::optional($event['scheduled_start_time'])->setTimezone(new DateTimeZone('UCT'))->format('Y-m-d\TH:i:s\Z'),
 				]
 			];
 		}
--- a/site/models/meet-events.php
+++ b/site/models/meet-events.php
@@ -27,7 +27,7 @@ class MeetEventsPage extends Page
 				'content'  => [
 					'title'    => $event['name'],
 					'link'     => $event['entity_metadata']['location'],
-					'date'     => Date::optional($event['scheduled_start_time'])->setTimezone(new DateTimeZone('UCT'))->format('Y-m-d\TH:i:s\Z'),
+					'date'     => $event['scheduled_start_time'],
 				]
 			];
 		}
--- a/site/snippets/templates/meet/events.php
+++ b/site/snippets/templates/meet/events.php
@@ -29,7 +29,7 @@
 <script>
 class LocalizedDatetimeElement extends HTMLElement {
 	connectedCallback() {
-		this.utc = new Date(this.date);
+		this.utc = new Date(this.getAttribute("date"));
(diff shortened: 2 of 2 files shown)


site/models/meet-events.php
```diff
@@ -28,8 +27,7 @@ public function children(): Pages
 				'content'  => [
 					'title'    => $event['name'],
 					'link'     => $event['entity_metadata']['location'],
-					'date'     => Date::optional($event['scheduled_start_time'])->setTimezone(new DateTimeZone($timezone))->format('Y-m-d H:i:s'),
-					'timezone' => $timezone
+					'date'     => Date::optional($event['scheduled_start_time'])->setTimezone(new DateTimeZone('UCT'))->format('Y-m-d\TH:i:s\Z'),
```
lukasbestle: `Typo fix` & `Suggestion`:

```suggestion
					'date'     => Date::optional($event['scheduled_start_time'])->setTimezone(new DateTimeZone('UTC'))->format('Y-m-d\TH:i:sp'),
```

However I think it would be even better to use a raw machine-parsable date format here and format it in the snippet (see below).

```suggestion
					'date'     => $event['scheduled_start_time'],
```

site/snippets/templates/meet/events.php
```diff
@@ -17,7 +17,7 @@ class="btn btn--filled"
 			<a href="<?= $event->link() ?>" class="flex items-center justify-between bg-white p-3 rounded shadow">
 				<h3 class="font-bold"><?= $event->title() ?></h3>
 				<p class="color-gray-700">
-					<?= $event->date()->toDate('D j M, H:i') ?> <?= $event->timezone() ?>
+					<localized-datetime><?= $event->date() ?></localized-datetime>
```
lukasbestle: `Suggestion` to keep the printed human-readable date in the HTML (= when the progressive enhancement fails) more readable:

```suggestion
					<localized-datetime date="<?= $event->date() ?>"><?= $event->date()->toDate('D j M, H:i T') ?></localized-datetime>
```

(Since our server uses the UTC timezone, we don't need to set the timezone explicitly. And even if it wasn't UTC, I think it would be fine as we will now have the visitor timezone.)

site/snippets/templates/meet/events.php
```diff
@@ -26,6 +26,30 @@ class="btn btn--filled"
 	<?php endif ?>
 </section>
 
+<script>
+class LocalizedDatetimeElement extends HTMLElement {
+  connectedCallback() {
+		this.utc = new Date(this.innerText);
```
lukasbestle:
```suggestion
		this.utc = new Date(this.date);
```

site/snippets/templates/meet/events.php
```diff
@@ -26,6 +26,30 @@ class="btn btn--filled"
 	<?php endif ?>
 </section>
 
+<script>
+class LocalizedDatetimeElement extends HTMLElement {
+  connectedCallback() {
```
lukasbestle: Formatting fix:

```suggestion
	connectedCallback() {
```

site/snippets/templates/meet/events.php
```diff
@@ -26,6 +26,30 @@ class="btn btn--filled"
...
+		this.utc = new Date(this.innerText);
+		this.innerText = this.format(this.utc);
+  }
```
lukasbestle: Formatting fix:

```suggestion
	}
```

PR #2328, opened 2024-06-08
Refactor KeyCDN thumb handling

lukasbestle:
### Summary of changes

- Refactor CDN plugin to use a separate `KeyCdn` class for image processing params
- Fixed use of image processing params according to KeyCDN docs
- Hook into Darkroom class for normalization
- Add support for focus point cropping and other Kirby thumb options
- Crop from the center by default (instead of the `smart` algorithm that tries to pick a focus point that may or may not match what we want)
- Remove `cdn()` helper that was only used internally in the CDN plugin

### Reasoning

- Cleaned up code that is easier to keep the overview on
- Increase reliability of image processing
- Increase consistency with our local GD/ImageMagick setups
- Allow us to use the full spectrum of Kirby thumb options in the future (we could add additional ones to the `KeyCdn` class if needed, I've started with the most common ones we already support with ImageMagick)

### Additional context

Relevant docs: https://www.keycdn.com/support/image-processing


distantnative (commented):
`Suggestion:` How about we use this chance to improve the naming a bit? I have found `cdn()` for the helper not that easy to understand that it's specifically image-related. Especially now that we use KeyCDN for more than image processing. Maybe the new class could rather be called something like `Kirby\Cdn\Image` , `Kirby\Cdn\Thumb` or so?

site/plugins/cdn/index.php
```diff
@@ -41,7 +41,7 @@
 			static $original;
 
 			if ($kirby->option('cdn', false) !== false) {
-				$url = cdn($file, $options);
+				$url = Image::url($file, $options);
```
distantnative: `Suggestion:` We could move this into `Kirby\Cdn\Fileversion` making sure that its URL always uses `Kirby\Cdn\Image::url()`.

site/plugins/cdn/src/Image.php
```diff
@@ -0,0 +1,135 @@
...
+	 * Generates a URL with KeyCDN image processing parameters
+	 * (https://www.keycdn.com/support/image-processing)
+	 * based on a file URL or object and Kirby thumb params
+	 */
+	public static function url(string|File $file, array $params = []): string
+	{
+		$kirby = App::instance();
+
+		$root = $file;
+		if (is_object($file) === true) {
```
distantnative:
```suggestion
		$kirby = App::instance();
		$root  = $file;

		if (is_object($file) === true) {
```

site/plugins/cdn/src/Image.php
```diff
@@ -0,0 +1,135 @@
...
+			$file = $file->mediaUrl();
+		}
+
+		$path = Url::path($file);
+
+		$darkroom = Darkroom::factory(
+			'im',
+			$kirby->option('thumbs', [])
+		);
+		$params = $darkroom->preprocess($root, $params);
```
distantnative: `Suggestion:` Move lines 39-43 to its own (maybe protected) method, e.g. `->options()` or `->params()`.

site/plugins/cdn/src/Image.php
```diff
@@ -0,0 +1,135 @@
...
+		$query = '';
+		if (empty($params) === false) {
+			$query = '?' . http_build_query([
+				...static::resizeOrCrop($params),
+				...static::grayscale($params),
+				...static::progressive($params),
+				...static::blur($params),
+				...static::sharpen($params),
+			]);
+		}
```
distantnative: `Suggestion:` Move lines 45-54 to its own (maybe protected) method, e.g. `->query()`.

site/plugins/cdn/src/Image.php
```diff
@@ -0,0 +1,135 @@
...
+
+		return $kirby->option('cdn.domain') . '/' . $path . $query;
+	}
+
+	protected static function blur(array $params): array
+	{
+		if ($params['blur'] !== false) {
+			$blur = max(0.3, min(100, $params['blur']));
+			return compact('blur');
```
distantnative:
```suggestion
			return [
				'blur' => max(0.3, min(100, $params['blur']))	
			];
```
`Reasoning:` I find `compact` always a bit harder to read/understand when glancing over code, especially when quickly trying to get the difference to the other return.

site/plugins/cdn/src/Image.php
```diff
@@ -0,0 +1,135 @@
...
+		};
+
+		return $query;
+	}
+
+	protected static function sharpen(array $params): array
+	{
+		if (is_int($params['sharpen']) === true) {
+			$sharpen = max(0, min(100, $params['sharpen']));
+			return compact('sharpen');
```
distantnative:
```suggestion
			return [
				'sharpen' => max(0, min(100, $params['sharpen']))
			];
```

`Reasoning`: Same as above.

site/plugins/cdn/index.php
lukasbestle: You mean the URL generation? Would mean that the constructor doesn't get the `url` prop but instead determines it from the original and modifications by calling `Image::url()`, right?

distantnative: Yes, that was my thought. Like

```php
	public function __construct(array $props)
	{
		parent::__construct([
			...$props,
			'url' => Image::url($props['original'], $props['modifications'])
		]);
	}
```

site/plugins/cdn/src/Image.php
lukasbestle: Hm, then we need another `$kirby = App::instance()` line. Also I think it adds too much boilerplate code without adding a benefit on its own. What are your thoughts behind the suggestion?

distantnative: Mainly making the `url()` method even easier to read. For understanding its main task, I'd only need to know that these are the options, not how the options are gathered via the Darkroom class. If I'd want to know that, I could look into the method that handles gathering/preprocessing the options.

It's a `Nitpick` though. 

I was earlier thinking, whether the `Image` class shouldn't be static, so it would rather be `(new Image($file, $options))->url()` then or so. That would avoid the repeated `App::instance()` calls. But also just a thought.

distantnative (commented):
I read through the code and couldn't spot any blocks, just one comment below. What I didn't check is if this fits with the KeyCDN API. I'd assume that's fine.

`Nitpick:` You've added all protected methods to the bottom of the class here. I just noticed this the other day also in the core with the `UpdateStatus` class. While @bastianallgeier and I follow a purely alphabetical order no matter whether public/protected/private (e.g. also in the recent Changes PRs https://github.com/getkirby/kirby/pull/6450/files#diff-882573b8078bdb642da7abafcffabb262e24a8ae7461916ef92890304bcb5859R125). I think it would be important to stick to one system. I've obviously a bias towards the pure-alphabetical (also cause it makes it easier to locate the method when you know the name). But I wouldn't mind that much either if we'd switch completely to your way of sorting them. I just think it'd be good to use one system for all of us.

lukasbestle:
About the method order: I agree it would be good to use a consistent pattern in the whole team. What I like about having `public` methods first is that it allows to see at a glance what I can do with a class. Whereas finding a specific method is easy with `Cmd-R` in VS Code and similar tools in other IDEs.

distantnative (approved):
As I said, I don't think it's a blocker, I think we can live with this inconsistency until we make a decision in the broader team.

PR #2329, opened 2024-06-08
Improved image sizing throughout the site

lukasbestle:
### Summary of changes

- Images loaded with the `img()` helper that are lazy-loaded (which is the default) now get a `sizes="auto"` attribute if not passed manually
- Most relevant LCP images "above the fold" are now loaded without lazy-loading
- `sizes` attribute is set for those LCP images so that the browser can immediately load the correct `srcset` variant

### Reasoning

- Increase load and render performance
- Reduce transfer sizes

### Additional context

Images "below the fold" are still lazy-loaded. With `sizes="auto"`, modern browsers can figure out the correct sizes better than we can manually. So I didn't touch these.


site/snippets/templates/home/playground/header.php
```diff
@@ -23,15 +23,18 @@ class="playground-header-figure-wrapper"
...
 						'srcset' => [
 							320,
 							640,
+							800,
 							960,
-							1280,
-							1600
+							1520,
+							1600,
+							3040
```
distantnative: `Thought:` 1600 seems a bit odd here as just 80px difference to 1520 - but then a large gap up to 3040.

site/snippets/templates/home/playground/header.php
```diff
@@ -23,15 +23,18 @@ class="playground-header-figure-wrapper"
 					<?= img($storyImage, [
 						'alt' => $storyImage->alt()->or('Panel screenshot for: ' . $story->title()),
 						'src' => [
-							'width' => 1280
+							'width' => 1520
 						],
 						'lazy' => false,
+						'sizes' => '(min-width: 1760px) 1520px, (min-width: 800px) 85vw, 100vw',
```
distantnative: `Question:` Is `85vw` here just a rough estimate or how did you arrive at this number?

distantnative (commented):
`Qustions:` Before I comment similar questions to many files, could you share some words on
- how you arrived at the specific `sizes` when set (if they are exact or just roughly right)
- How you have picked the `srcset` widths

Thank you! :)

site/snippets/templates/home/playground/header.php
lukasbestle: `1520` and its 2x counterpart `3040` are the exact sizes for when the browser width is larger than our layout max-width.

The `1600` came in as a 2x counterpart of `800`. But you are right that it is probably not needed because the browser might as well choose `1520` then.

Maybe we should rather redistribute all other sizes except `1520` and `3040` with the goal to reduce large gaps.

site/snippets/templates/home/playground/header.php
lukasbestle: See general comment below

lukasbestle:
In general: The fluid layout makes everything a sliding scale. So:

- the `srcset` numbers are mostly chosen by feeling except those exact sizes in the `max-width` case we know,
- the `sizes` thresholds are also rough as I couldn't reliably determine all exact breakpoints (if you know or can determine them, feel free to edit those directly) and
- the `sizes` vw units are rough by definition because of the fluid layout (the margins are often static, but the images scale fluidly, which prevents the ratios from being exact for the entire scale). I have chosen those to fit for the part of the breakpoint where the image is 80-90% of its maximum size.

site/snippets/templates/home/playground/header.php
distantnative: Sounds good to me. Also for the other changes. If we can keep those "must-have" sizes and then rather redistribute a few alternatives as "evenly" as possible.

site/snippets/templates/buy/good-cause.php
```diff
@@ -29,8 +29,8 @@
 									'width' => 300
 								],
 								'srcset' => [
-									'1x' => 400,
-									'2x' => 800,
```
distantnative, line 33: `Question:` You wrote `1x`/`2x` aren't needed/used together with `sizes`. But given that no sizes are passed here, wouldn't it still help for all browsers not supporting `sizes="auto`?

content/releases/4-0/1_design/snippet.php
```diff
@@ -36,7 +36,20 @@
 </style>
 
 <figure class="release-box bg-light mb-6">
-	<?= $section->image('company.png')->resize(1600) ?>
+	<?= img($section->image('company.png'), [
+		'src' => [
+			'width' => 1248
+		],
+		'lazy' => false,
+		'sizes' => '(min-width: 1540px) 1248px, (min-width: 1160px) calc(77.78vw + 66px), (min-width: 480px) calc(100vw - 96px), 90vw',
```
distantnative, line 45: `Comment:` These are by nature to be trusted (or using the same tool, which I just assume will generate the same results for me), but there is no easy natural way to derive at these values during review. I am wondering if we should add a comment where these are coming from for a few years down the road when we're wondering. Although it might be a bit annoying to add the comment everywhere. But can't think of a good central place.

site/snippets/templates/buy/good-cause.php
lukasbestle: Yes, if the browser doesn't support `sizes="auto"`, the widths won't be useful. However I feel like the `2x` size at the moment is also not useful as none of our images render at a fixed size. A 1x device may still need a larger version while a 2x device could end up needing the smaller one. Best way would be to add `sizes` manually everywhere, but I don't think it's worth it for the transition. The current changes are already an improvement over the status quo.

The issue with `1x`/`2x` is that it's plain invalid when used together with `sizes`. The browser needs to know the actual image widths to be able to choose one by its rendered size and resolution.

distantnative (approved):

PR #2332, opened 2024-06-13
4.3 docs

distantnative:


bastianallgeier:
I'm afraid, the Docs section in Notion was pretty rough. We probably should also add docs for: 

- `disablepictureinpicture` here https://getkirby.com/docs/reference/text/kirbytags/video
- Plugin registration with named arguments

distantnative:
I left the plugin registration out, I was afraid that it causes quite some confusion etc. So thought about making the switch in the docs only in v5?

bastianallgeier (approved):

Issue #1289, opened 2021-04-15
Disable accessibility on `playground-browser`


distantnative:
Somehow the `aria-hidden`/`role=presentation` got lost again, which is why it's throwing an accessibility warning for this element (while it should actually be ignored).

bastianallgeier:
It should have a role=img

distantnative:
Seems like this isn't sufficient:
<img width="757" alt="Screen Shot 2021-04-16 at 18 45 03" src="https://user-images.githubusercontent.com/3788865/115057066-e152d900-9ee3-11eb-8e74-e4954c90b009.png">

bastianallgeier:
Maybe we can ask @sebastiangreger about his opinion. I don't see how we could improve this any more. The headline is meant to be a part of a visual and not to be read out loud. We already mark it properly as far as I could find out in the HTML. I just don't know how to ignore the contrast issue. If this was an actual img with the same contrast issues, lighthouse would accept it.

sebastiangreger:
Always happy to nerd out over some delicious semantics :wink:

I don't have experience with this particular context of a figure consisting of HTML elements, but my hunch is that `role="img"` is the culprit here.

Lighthouse checking all elements within the tag's content actually makes sense. Displaying an image with essential information where the text does not meet the contrast criteria would be a violation of [WCAG criterion 1.4.11](https://www.w3.org/TR/WCAG21/#non-text-contrast). It's just that automated testing can't evaluate that for bitmaps (so e.g. for an important infographic PNG with poor contrast, LH might award a 100% score while a human reviewer would flag a WCAG violation).

![a11yexplorer](https://user-images.githubusercontent.com/6355217/115397616-963c0d00-a1e6-11eb-9405-c108c0015aca.png)

So `role="img"` does not make LH skip the element's content. Since there are no other indicators for assistive tech to ignore this "graphic" (as it's represented in the accessibility tree), it gets evaluated ...and subsequently fails the test. We cannot really add `aria-hidden="true"` or `role="presentation"` to the `<figure>` element itself, as that would misrepresent the content to an AT user (we would not have an alternative representation of non-text content; [WCAG 1.1.1](https://www.w3.org/TR/WCAG21/#non-text-content) failure).

Since this particular graphic is illustrational to the degree that reading the text in it is not "required for understanding" (quote WCAG re when contrast requirements apply for images), we'd need to communicate that to the assistive tech, so LH can ignore it.

[...]

PR #2340, opened 2024-06-22
Hide presentational content from a11y technology

lukasbestle:
### Summary of changes

Implement `aria-hidden` as suggested and discussed in https://github.com/getkirby/getkirby.com/issues/1289.

### Reasoning

- Avoid announcing purely presentational content to a11y technology (there's already a much more useful `aria-label`)
- If we are lucky, this should also make the PageSpeed test happy

### Additional context

As Lighthouse in Chrome didn't complain about the font sizes (only the PageSpeed test), I couldn't verify this yet. But even if PageSpeed still complains, this is an improvement for users of a11y tech.


distantnative:
I just tried it with VoiceOver on Mac and there are some observations:

- `Review`: Moving through the elements with `Control + Option + Arrow` seems to work well. It does announce "empty figure" and "empty group" but that seems ok to me.
- `Todo`: Jumping around focus with tab prdocues some inconsistent results: 
  - The filesystem items get focussed. I don't think that makes a lot of sense for screenreader users when in the normal flow this just gets read out as one group item.
  - The example API response receives focus. In this context weird as it is the only example part.
- `Todo`: The PHP and Vue logos have no text read out.
- `Suggestion`: The descriptions read out seem to me a bit disconnected. E.g. `A browser window with an example website` on its own without the visual leaves me a bit lost what this is supposed to tell me. Maybe we need to think making this description a bit mroe illustrative and matching the purpose the figure has, e.g. `A browser window showing a website created with Kirby`

lukasbestle:
Implemented most of your observations and suggestions. For the focus issues I don't have a good impromptu solution. Maybe we could cover that in a separate PR.

distantnative:
PHP/Vue only read `image` - not sure what VoiceOver needs to read out the title.

distantnative (approved):
Yes, we can improve this in another PR since that issue existed before.

PR #2342, opened 2024-06-23
Privacy-friendly video embeds

lukasbestle:
### Summary of changes

- New custom `video()` helper that only displays the cover by default and loads the `<iframe>` on click
- Add covers to content folder to avoid requests to YouTube's servers without consent
- Embed `<iframe>` with `no-referrer` policy
- No longer disable player controls to improve UX
- Add YouTube to our privacy policy

### Reasoning

- Increases visitor privacy
- Increases performance
- Allows to easily open videos in a new tab or copy links as the cover is linked with a plain video link


site/models/buzz-entry.php
```diff
@@ -32,20 +32,14 @@ public function metadata(): array
 
 	public function url($options = null): string
 	{
+		if ($this->video()->isNotEmpty() === true) {
+			return $this->video()->value();
+		}
```
distantnative: `Todo`: We recently just changed it, so that videos (e.g. release stream) can have their own buzz entry page to link to (sending people to our website rather than to YouTube). With this change, those pages cannot be reached anymore (e.g. also preventing that we could some longer description alongside the video). I'd vote to revert this (and related) changes.

site/plugins/site/helpers/html.php
```diff
@@ -119,6 +122,46 @@ function version(string $version, string $format = '%s'): string
...
+ */
+function video(
+	string $url,
+	File $cover,
+	array $options = [],
+	array $attr = [],
+): string|null {
+	if (Str::contains($url, 'vimeo', true) === true) {
+		throw new Exception('Vimeo support is not implemented');
+	}
```
distantnative, line 135:
```suggestion
```

`Question`: The core implementation can also take care of Vimeo, can't it?

site/plugins/site/snippets/video.php
```diff
@@ -0,0 +1,28 @@
...
+
+		<svg version="1.1" viewBox="0 0 68 48" class="play-button">
+			<path class="ytp-large-play-button-bg" d="M66.52,7.74c-0.78-2.93-2.49-5.41-5.42-6.19C55.79,.13,34,0,34,0S12.21,.13,6.9,1.55 C3.97,2.33,2.27,4.81,1.48,7.74C0.06,13.05,0,24,0,24s0.06,10.95,1.48,16.26c0.78,2.93,2.49,5.41,5.42,6.19 C12.21,47.87,34,48,34,48s21.79-0.13,27.1-1.55c2.93-0.78,4.64-3.26,5.42-6.19C67.94,34.95,68,24,68,24S67.94,13.05,66.52,7.74z" fill="#f00"/>
+			<path d="M 45,24 27,14 27,34" fill="#fff"/>
+		</svg>
+	</a>
+	<span class="text-overlay">
+		This video is loaded from YouTube servers with your consent.
+		<a href="<?= url('privacy#website__third-party-services') ?>" class="link">Read more…</a>
+	</span>
```
distantnative: `Praise:` I like how the styling looks of this overlay text!

assets/js/components/video.js
```diff
@@ -0,0 +1,12 @@
+export default class {
+	constructor() {
+		document.querySelectorAll(".video-embed > a").forEach((element) => {
+			element.addEventListener("click", this.click);
+		});
+	}
+
+	click(e) {
+		e.preventDefault();
+		e.currentTarget.parentElement.outerHTML = e.currentTarget.dataset.iframe;
```
distantnative, line 10: `Suggestion:` I was wondering if we want to hide any of the overlay texts once one click was taken (consent given)?

site/plugins/site/extensions/tags.php
```diff
@@ -48,16 +48,18 @@
...
 $tags['screencast'] = [
 	'attr' => [
+		'cover',
 		'title',
 		'text',
 	],
 	'html' => function ($tag) {
 		return snippet('kirbytext/screencast', [
 			'url'   => $tag->value,
+			'cover' => $tag->file($tag->cover),
```
distantnative:
```suggestion
			'cover' => $tag->cover ? $tag->file($tag->cover) : $tag->file('youtube.jpg'),
```

`Suggestion:` We could keep our rest of the code DRY-er if we add this default.

site/plugins/site/helpers/html.php
```diff
@@ -119,6 +122,46 @@ function version(string $version, string $format = '%s'): string
...
+	$url = str_replace('www.youtube-nocookie.com', 'www.youtube.com', $url);
+
+	// but always use privacy mode for embedding
+	$privacyUrl = str_replace('www.youtube.com', 'www.youtube-nocookie.com', $url);
+
+	// get the iframe string from the core
+	$iframe = Html::youtube($privacyUrl, [
+		...$options['youtube'] ?? [],
+		'autoplay'       => 1,
+		'modestbranding' => 1,
```
distantnative:
```suggestion
```

Has been deprecated and won't have any effect: https://developers.google.com/youtube/player_parameters#release_notes_08_15_2023

site/plugins/site/helpers/html.php
```diff
@@ -119,6 +122,46 @@ function version(string $version, string $format = '%s'): string
...
+		'showinfo'       => 0,
+		'rel'            => 0,
+	], [
+		...$attr,
+		'referrerpolicy' => 'no-referrer'
+	]);
+
+	return Tpl::load(dirname(__DIR__) . '/snippets/video.php', compact('attr', 'iframe', 'url', 'cover'));
```
distantnative:
```suggestion
	return  snippet('video', compact('attr', 'iframe', 'url', 'cover'), true);
```

`Suggestion`: I would not start splitting up our snippets between `site/snippets/` and plugins. That will make it just harder later on to find the code. If we place it with the regular snippets, we can also use the `snippet()` helper here.

site/plugins/site/helpers/html.php
lukasbestle: Currently we don't use Vimeo embeds as far as I can see. If we ever need them, we also need to update the privacy policy, the overlay text etc. It's not just about the iframe.

site/models/buzz-entry.php
lukasbestle: Ah, right. Didn't realize the links to the buzz pages were on purpose.

Reverted it. What are related changes?

assets/js/components/video.js
lukasbestle: Because we replace the `outerHTML` of the whole container, the overlay is completely removed from DOM.

distantnative: Yes, my suggestion is to additionally remove the text from all other video embeds.

lukasbestle: I'm confused...

lukasbestle: Ah, you mean once a user consents to a any video, all other videos will load by default?

distantnative: Not quite, because if we load them all, the buzz overview page has multiple videos autoplaying at once. That's why I thought, we only hide the text overlay for these others, so they look like a normal video with just the play button.

site/models/buzz-entry.php
distantnative: Thought that maybe that part was also taken out of the template. But all good.

assets/js/components/video.js
lukasbestle: I removed `autoplay` together with the `sessionStorage` change for that reason. The `autoplay` attribute doesn't seem to be working reliably anyway as YouTube often ignores it because it cannot reliably detect if user interaction was involved.

I have now tried it as you suggest to only remove the text (see last commit), but the end result is confusing IMO. Without the text, you don't see why you need to click twice if `autoplay` doesn't work. Also it causes flickering when switching pages.

But we have three implementations to choose from now. TBH I'd vote for the first implementation with no `sessionStorage` at all. It works most reliably.

distantnative: I like the newest one, but for me the autoplay also works always (which then made all videos play on the buzz overview with the second implementation). So for me the order would be

- remove all text overlays
-  initial implementation
- remove all overlays

If you don't like the text overlay variant or, rather, the autoplay doesn't reliably work, let's go with the initial one.

lukasbestle: TBH I can't get any of my browsers (Safari, Chrome, Firefox) to autoplay with the second version. Also I removed the `autoplay` attribute there, so it's even more weird.

lukasbestle: With the third version I can sometimes get autoplay to work. In Firefox it's quite reliable, in Safari it works sometimes, but in Chrome it never works for me. And without autoplay, I think the removed text is quite confusing.

distantnative: Let's go with the initial version then.

distantnative (approved):

PR #2349, opened 2024-07-08
Partner leads form

distantnative:
### Summary of changes
- Replace Airtable leads form with our own (similar to partner applications)
- Send email on form submit to ourselves with the details from the form 
- Add email SMTP config for local and Fastmail
- Move spam protection logic to `Honey\Time` class/plugin to keep code DRY
- Add dropdown arrow for select inputs in `.dialog-form`
- Add CSS wrapper `.autogrow` for auto-sizing textareas

Screenshot 2024-07-12 at 08 59 15


bastianallgeier:
It already looks pretty great to me. I would use a separate success page, like we do for the partner signup form. Especially, since we send out emails. 

I also wonder if we need to ramp up anti-spam checks even more for this form. Sending out emails in combination with form data is quite scary.

distantnative:
@bastianallgeier ok for the separate success page.

On spam protection, I'm open for known good practices.

The good thing is that most of the form data only goes into the email to us where the recipient (we) is hardcoded. Wondering if we should eliminate the name variable from the email to the customer as well, less personal but also then that email has no entry point for spam messages.

bastianallgeier (approved):
I'm really happy with the state of the form now

PR #2351, opened 2024-07-16
Update tc() Reference

tobiasfabian:
## Description
Document [`tc()`](https://getkirby.com/docs/reference/templates/helpers/tc) helper method. You can define an array for the translation key, to return different strings depending on `$count`

### Additional context

https://getkirby.com/docs/reference/objects/toolkit/i18n/translate-count


bastianallgeier (approved):

bastianallgeier:
Much better! Thank you for the improvement.

PR #2353, opened 2024-07-26
Update certified partner profile

bezin:
Hi all,

we updated our partner profile. The existing gallery-items all have a UUID, however UUIDs are disabled in the config. Do I have to provide UUIDs? If yes, what is the best way as the CLI commands won't work. 

Best
Benedict


texnixe:
The UUIDs are not used, they are just automatically created by the import script I run to fetch the profile data from Airtable.

PR #2354, opened 2024-07-30
Update partner card image and summary

bezin:
Forgot to add summary and new image with 2/1 aspect ratio in my last PR.


distantnative:
@bezin 2:1 doesn't replace the card pic though (which is still used for the main profile page), it's called "stripe" instead.

bezin:
Thanks for the hint. It worked as expected locally, but that was a caching issue. Restored the original card.jpg and added a stripe.jpg.

PR #2355, opened 2024-08-06
Scenario page: education

distantnative:
## Description

- [x] Might need a Panel screenshots/mockups or so to show off the CMS itself
- [x] Images get squeezed when resizing

### Summary of changes
- Adds a new type of pages: scenario-/industry-specific landing pages
- First scenario page: education
- Moves all features pages also into `for`, e.g. `getirby.com/for/developers`


bastianallgeier (approved):
I really like it a lot! We should think about creating a better dropdown menu for when we add more of those "For" pages. But as a first step to get started with additional pages, this is really great.

PR #2356, opened 2024-08-07
`For hospitality` scenario page

distantnative:
## Description

- [x] First merge https://github.com/getkirby/getkirby.com/pull/2355

### Summary of changes
- New scenario page for hospitality (restaurant, bars, traveling agencies, tour operators, hotels...)


distantnative:
@bastianallgeier so far only adding footer links for those pages - I don't yet think they should be on the main menu, but I guess for SEO they need to be linked somewhere. However, the footer menu also starts looking a bit overloaded.

PR #2359, opened 2024-08-13
fix docs: replace tabs with spaces in yaml code examples

rasteiner:
## Description
I've replaced ~~all tabs~~ the tabs I've found with spaces in yaml code examples. 

### Reasoning
Tabs are explicitly discouraged by the spec:
https://yaml.org/spec/1.2.2/#61-indentation-spaces

Also, there were multiple instances where tabs and spaces were mixed in examples, and since the tabs are displayed as four spaces in the docs, it looked like there were wrong indentations:
image


bastianallgeier (approved):
This is super helpful! Thank you <3

PR #2360, opened 2024-08-13
Consistent license hub links

distantnative:
- Added license hub links to license terms
- Using `hub.getkirby.com` throughout site


bastianallgeier:
What's your reasoning behind the changes? I thought we decided to go for hub.getkirby.com a while ago.

distantnative:
We can also go for the hub one - just think it should be one consistent URL and not both in different places.

Couldn't remember that decision. To me licenses felt more self explanatory also for users later.

bastianallgeier:
@distantnative I agree that it should be consistent. I think the general idea was that the hub could and should go beyond just licenses later. That's why we chose hub instead of licenses.

distantnative:
@bastianallgeier changed it

bastianallgeier (approved):

PR #2373, opened 2024-10-03
`For artists` page

distantnative:


bastianallgeier (approved):
Love it!

bastianallgeier:
We could add https://axelscheffler.de and https://www.andreasgursky.com

PR #2379, opened 2024-10-18
Added a hint that auth.challenge.email config is also used for pw-reset

spackmat:
## Description

It took me a moment to find the code creating the actual email for the password-reset and to realize that it also uses the email config from `auth.challenge.email` as the reset is also a challenge from that point of view. 

### Summary of changes

This commit drops a hopefully helpful note on that.

### Reasoning

From reading the docs I expected the `password-reset` functionality to use its own email configuration within the `auth` config or its own email preset, but could not find any information on that.

### Additional context

None.


bastianallgeier (approved):

bastianallgeier:
Thanks for the addition!

PR #2382, opened 2024-11-21
Switch donations to Against Malaria Foundation

distantnative:


lukasbestle (approved):
We need to make the same config changes in the hub afterwards.

PR #2389, opened 2024-12-07
Update PHP EOL

distantnative:
According to https://www.php.net/supported-versions.php  - I don't know if they've extended them that they're now a full year after we noted them.

Also added PHP 8.4.

Do we need to update `tests/Cms/System/UpdateStatusTest.php` simultaneously with merging this PR?


lukasbestle (approved):
Thanks, good catch!

The tests are completely mocked, so I think they should all still pass. Of course the dates in the mock files are now also outdated, but IMO that's not an issue as we can still test the logic.

PR #2397, opened 2025-01-15
Add plugins to algolia index with upgraded the dependency to v4

afbora:
## Description

- [x] Algolia dependency upgraded to v4 api
- [x] Codes adapted to v4 api
- [x] Added the new collection named `plugins`
- [x] Plugins are injected to the algolia search index
- [x] Set algolia index as `getkirby-4`


site/plugins/search/src/Entry.php
```diff
@@ -128,18 +128,22 @@ protected function templates(): array
 	 */
 	public function toData(): array
 	{
-		$data = ['objectID' => $this->page->id()];
+		$data             = [];
+		$data['objectID'] = match ($this->template()) {
+			'plugin' => $this->page->url(),
+			default  => $this->page->id()
+		};
```
distantnative: `Suggestion:` I would keep `objectID` always with an id and for all entries also adding the URL as `url` separately. I think that'll be helpful long term. We only need to change the frontend part then to use `url`.

afbora: I've reverted the change. Btw `/plugins/developer/plugin-name` uri automaticly already redirected to new subdomain 🎉 I wasn't aware it.

https://getkirby.com/plugins/tobimori/seo

So no need to handle from JS side.

distantnative:
Seems to work well for me!

Will we add a cronjob to flush the search index regularly? @bastianallgeier or can we set up ploi that also when we push to plugins.getkirby.com it will flush the search index of getkirby.com?

afbora:
Yes, content changes on getkirby.com are tracked by the algolia github workflow, but it should also be triggered when a plugin is added to plugins.getkirby.com.

Maybe we can trigger getkirby.com algolia workflow when plugins repository updated. I've found an article about that:
https://medium.com/hostspaceng/triggering-workflows-in-another-repository-with-github-actions-4f581f8e0ceb

bastianallgeier:
We could easily flush the cache from a plugins.getkirby.com deployment for getkirby.com via a webhook. Should not be a problem at all.

PR #2398, opened 2025-01-16
Send Discord webhook when a new partner signs up 

bastianallgeier:


distantnative:
`Suggestion:` How about we add some rules to our `.gitignore` to not include README etc. from plugins in the repo?

site/plugins/gravatar/index.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+function gravatar(string $email, string|null $default = null): string {
+	$address = strtolower(trim($email));
+	$hash    = hash('sha256',$address);
+
+	return 'https://www.gravatar.com/avatar/' . $hash . '?d=' . $default;
+}
```
distantnative, line 8: `Suggestion:` Don't we maybe want to make this part of the discord plugin as we would need it in multiple repos and maybe also helpful for others?

bastianallgeier: It felt a bit out of scope. I was wondering if it should be its own super small official plugin that can be used in multiple other places as well. But on the other hand I really also don't mind too much. Could be in the Author class as well.

bastianallgeier:
@distantnative ~~good idea.~~ Thought about it again. To a certain extend, the site repo is serving as inspiration for other users and maybe it's a good idea to keep the plugin readmes in there for them, even if we don't need them.

distantnative (approved):
I would add the gravatar feature to the plugin. But we can also merge this first and change it when we add that in v1.2 of the plugin/

bastianallgeier:
I think we can keep it as it is and then add gravatar support later. We should probably also use the plugin on the other sites that send Discord webhooks in the future. I just didn't want to spend a lot more time on it.

PR #2404, opened 2025-02-05
Update "Alternative multisite setup" for correct visual folder structure

3tlam:
Heya all

This is my first ever PR; hope I'm doing it correctly 

The previous folder structure made it look, as if `my-site.com` & `my-shop.com` were nested within the `kirby` folder, whereas they should be next to each other. Removing the spacing sould solve this visually


bastianallgeier (approved):
Thank you very much for your first PR and for fixing the indentation issue 🎉

PR #2405, opened 2025-02-06
Add meet.json.php template

texnixe:

### DIFF
--- a/site/templates/meet.json.php
+++ b/site/templates/meet.json.php
@@ -1,40 +1,33 @@
 <?php
 /**
- * @var Kirby\Cms\Pages $upcoming
- * @var Kirby\Cms\Pages $past
+ * @var \Kirby\Cms\Pages $upcoming
+ * @var \Kirby\Cms\Pages $past
  */
-$upcomingEvents = [];
-$pastEvents     = [];
 
 $since = param('since');
 
-foreach ($upcoming as $event) {
-	$upcomingEvents[] = [
-		'title'    => $event->shortTitle()->value(),
-		'date'     => $event->datetime()->format('c'),
-		'timezone' => $event->timezone()->value(),
-		'link'     => $event->link()->value(),
-		'isMeetup' => $event->isMeetup(),
-	
-	];
-}
+$upcomingEvents = $upcoming->toArray(fn($event) => [
+	'title'    => $event->shortTitle()->value(),
+	'date'     => $event->datetime()->format('c'),
+	'timezone' => $event->timezone()->value(),
+	'link'     => $event->link()->value(),
+	'isMeetup' => $event->isMeetup(),
+]);
 
 // Only meetups, no other events
 $past = $past->filterBy('city', '!=', '');
 
 // Use param since to filter meetups after given date
-if ($since = param('since')) {
+if (is_string($since)) {
 	$past = $past->filterBy('num', '>', $since);
 }
 
-foreach ($past as $event) {
-	$pastEvents[] = [
-		'title' => $event->shortTitle()->value(),
-		'date'  => $event->datetime()->format('c'),
-		'link'  => $event->link()->value(),
-		'isMeetup' => $event->isMeetup(),
-	];
-}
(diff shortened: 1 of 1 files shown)


site/templates/meet.json.php
```diff
@@ -0,0 +1,42 @@
...
+foreach ($upcoming as $event) {
+	$upcomingEvents[] = [
+		'title'    => $event->shortTitle()->value(),
+		'date'     => $event->datetime()->format('c'),
+		'timezone' => $event->timezone()->value(),
+		'link'     => $event->link()->value(),
+		'isMeetup' => $event->isMeetup(),
+	
+	];
+}
```
lukasbestle: `Suggestion`: This could be shortened with:

```suggestion
$upcomingEvents = $upcoming->toArray(fn ($event) => [
	'title'    => $event->shortTitle()->value(),
	'date'     => $event->datetime()->format('c'),
	'timezone' => $event->timezone()->value(),
	'link'     => $event->link()->value(),
	'isMeetup' => $event->isMeetup(),
]);
```

site/templates/meet.json.php
```diff
@@ -0,0 +1,42 @@
...
+
+foreach ($past as $event) {
+	$pastEvents[] = [
+		'title' => $event->shortTitle()->value(),
+		'date'  => $event->datetime()->format('c'),
+		'link'  => $event->link()->value(),
+		'isMeetup' => $event->isMeetup(),
+	];
+}
```
lukasbestle:
```suggestion
$pastEvents = $past->toArray(fn ($event) => [
	'title'    => $event->shortTitle()->value(),
	'date'     => $event->datetime()->format('c'),
	'link'     => $event->link()->value(),
	'isMeetup' => $event->isMeetup(),
]);
```

site/templates/meet.json.php
```diff
@@ -0,0 +1,42 @@
...
+		'isMeetup' => $event->isMeetup(),
+	
+	];
+}
+
+// Only meetups, no other events
+$past = $past->filterBy('city', '!=', '');
+
+// Use param since to filter meetups after given date
+if ($since = param('since')) {
```
lukasbestle: `Cleanup`: This variable is already set in line 9. We can use this to improve type safety:

```suggestion
if (is_string($since) === true) {
```

site/templates/meet.json.php
```diff
@@ -0,0 +1,42 @@
...
+		'date'  => $event->datetime()->format('c'),
+		'link'  => $event->link()->value(),
+		'isMeetup' => $event->isMeetup(),
+	];
+}
+
+echo json([
+	'upcoming' => $upcomingEvents,
+	'past'     => $pastEvents,
+]);
```
lukasbestle, line 35: `Suggestion`: If we reuse the `$upcoming` and `$past` variables instead of setting the arrays to `$*Events` variables, we could do:

```suggestion
echo json(compact('upcoming', 'past'));
```

site/templates/meet.json.php
```diff
@@ -0,0 +1,42 @@
+<?php
+/**
+ * @var Kirby\Cms\Pages $upcoming
+ * @var Kirby\Cms\Pages $past
```
lukasbestle: `Cleanup`: Fully-qualified class names improve compatibility with IDE tools:

```suggestion
 * @var \Kirby\Cms\Pages $upcoming
 * @var \Kirby\Cms\Pages $past
```

lukasbestle (commented):
It's great and already quite clean. A few suggestions about code quality, totally optional:

site/templates/meet.json.php
distantnative: Or `->values(fn () => ...)` if you don't want keys.

site/templates/meet.json.php
texnixe: If I do that, my IDE tells me to replace qualifiers with an import...

site/templates/meet.json.php
texnixe: Not a fan of compact, so never use it

PR #2409, opened 2025-02-10
5.0-beta.3

distantnative:


lukasbestle (commented):
Awesome! There seem to be some enhancements in Notion that haven't made it, is this intended?

distantnative:
They felt very "internal", trying to weigh the utility vs. making the list longer. Maybe I landed on the wrong decision?

lukasbestle:
You are right about that. I'll leave it to Basti.

bastianallgeier:
I've added a bit of text for the batch delete docs page and mentioned the content representation support for the site controller in the docs. But the rest looks great to me.

content/releases/5/3_entries-field/snippet.php
```diff
@@ -0,0 +1,49 @@
...
+</style>
+
+<div class="v5-entries-columns columns">
+	<div class="release-text-box">
+		<?php snippet('templates/release-4/teaser', ['section' => $section]) ?>
+	</div>
+	<figure class="v5-entries-hero release-padded-box bg-light grid place-items-center">
+		<?php snippet('templates/release-4/image', [
+			'alt' => 'The new entries field allows you to create and manage multiple entries for the same field type',
+			'img' => $section->image('entries-field.png')->resize(1500)
```
distantnative, line 43: Not sure if that's a good size.

content/releases/5/4_batch-delete/snippet.php
```diff
@@ -0,0 +1,33 @@
...
+			"teaser code screenshot"
+	}
+}
+</style>
+
+<div class="v5-batch-columns">
+	<figure class="release-box bg-light grid place-items-center" style="grid-area: screenshot">
+		<?php snippet('templates/release-4/image', [
+			'alt' => '',
+			'img' => $section->image('batch-delete.png')->resize(850)
```
distantnative: Needs an actual alt text. And also not sure about the size here either.

content/releases/5/4_batch-delete/snippet.php
```diff
@@ -0,0 +1,33 @@
...
+	gap: var(--spacing-6);
+	grid-template-areas:
+		"screenshot"
+		"teaser"
+		"code"
+}
+
+@media screen and (min-width: 80rem) {
+	.v5-batch-columns {
+		grid-template-columns: 1fr 1fr 2fr;
```
distantnative: Leaves the teaser and code boxes rather large with little content. Didn't find any good setup yesterday with the rather square image.

distantnative (commented):
Added comments where I think it's still a little rough.

content/releases/5/3_entries-field/snippet.php
bastianallgeier: Looks good to me on regular and retina displays

content/releases/5/4_batch-delete/snippet.php
bastianallgeier: I've added the alt text. Size looks good.

content/releases/5/4_batch-delete/snippet.php
bastianallgeier: I think this is fine. I've moved the code example to the right, which makes the layout a bit more interesting in my opinion.

PR #2421, opened 2025-03-06
A11yn `icon()` helper

distantnative:
### Summary of changes
- `icon()` accepts a new optional `$title` argument
- If not set (and SVG doesn't natively include a `<title>` tag), `aria-hidden="true"` is added to the SVG
- If set, `role="img"` as well as `aria-labelledby` are added to the SVG and a corresponding `<title>` tag inserted as first child


lukasbestle (commented):
Looks great!

I went over all uses of the `icon()` helper using code search and added the titles or other kinds of accessible labels where the icons are not just presentational. I don't have much experience with testing the accessibility tree though, so I'd be grateful for tests and input.

distantnative:
That's great, thank you! Where I am a bit unsure is this use for the loading spinner icon. I think for that `role="alert"` would be more appropriate. And I've see this combined with `aria-live`. Sometimes even `aria-busy`.

lukasbestle:
Is it really an alert that needs attention? `aria-busy` sounds like it could match what we need.

lukasbestle:
Just noticed that `/docs/reference` is broken with:

```
String could not be parsed as XML

SimpleXMLElement::__construct(): Entity: line 4: parser error : Opening and ending tag mismatch: symbol line 2 and svg (2)
```

It's in `site/snippets/templates/reference/entry/decoration.php:3`, but not sure which icon is affected.

assets/icons/bluesky.svg
lukasbestle, line 1: The Bluesky icon from the Panel set is just an outline. This looks weird compared to all the other social icons:

Bildschirmfoto 2025-03-09 um 17 09 30

lukasbestle:
Apart from that I think we can merge this as well.

assets/icons/bluesky.svg
distantnative: Didn't realize the difference. Added the filled version back in.

PR #2424, opened 2025-03-11
Fix link Update reference-article.txt

Sean-OShea:
## Description
A typo in the link made the href fall back to "https://getkirby.com" instead of:

https://getkirby.com/docs/guide/configuration#the-config-php__using-multiple-config-files

### DIFF
--- a/content/docs/2_reference/6_system/1_options/0_debug/reference-article.txt
+++ b/content/docs/2_reference/6_system/1_options/0_debug/reference-article.txt
@@ -25,7 +25,7 @@ In a production environment, always log errors to your PHP error logs.
 
 To avoid accidentally enabling debugging in production, but to still be able to use it in certain situations, you have different options:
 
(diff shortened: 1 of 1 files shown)


Sean-OShea:
The fix aside, i would suggest to link to another part of the page that i think would be more relevant to the topic of configuring a multi-environment setup: https://getkirby.com/docs/guide/configuration#multi-environment-setup

At least this is what i was looking for when looking this up.

bastianallgeier:
@Sean-OShea you are totally right. It should link to the page you suggested. Can you change it in the PR? Otherwise I will take care of it.

Sean-OShea:
> @Sean-OShea you are totally right. It should link to the page you suggested. Can you change it in the PR? Otherwise I will take care of it.

Thanks for the confirmation, it's done.

bastianallgeier (approved):

bastianallgeier:
Perfect, thank you for the fix!

PR #2432, opened 2025-03-18
[v5] fix: Improved support for DocBlock parameters

### LINKED ISSUE #2431: Reference: Respect custom `@param` definitions for splat param
E.g. https://getkirby.test/docs/reference/objects/toolkit/collection/sort displays the params as `...$args`, which is taken from the actual method params (at least for v5).

However the core docblock specifies more specific params. We could use those instead of a splat param:

https://github.com/getkirby/kirby/blob/e29f216630ee1c2ab7e3f8f1bb3d335541121587/src/Toolkit/Collection.php#L995-L1003

distantnative:
### Summary of changes
- DocBlock parameters are now also considered if they describe additional parameters not picked up by the PHP reflection
- `$args` parameter is removed if any other parameter is documented
- `null` as default value for parameters is now displayed

### DIFF
--- a/site/plugins/site/src/Reference/Reflectable/Tags/Parameter.php
+++ b/site/plugins/site/src/Reference/Reflectable/Tags/Parameter.php
@@ -30,13 +30,14 @@ class Parameter
 	}
 
 	public static function factory(
-		ReflectionParameter $parameter,
+		ReflectionParameter|null $parameter = null,
 		ParamTagValueNode|null $doc = null,
 		Reflectable|null $context = null
 	): static {
-		$name    = $parameter->getName();
+		$name    = $parameter?->getName();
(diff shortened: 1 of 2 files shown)


bastianallgeier:
Is there a good place where I can check the before and after state?

distantnative:
https://getkirby.test/docs/reference/objects/toolkit/collection/sort

But @lukasbestle wanted to review this, so you might not have to do it.

bastianallgeier:
There seems to be an issue with optional parameters that have null as default value. 

Check out the second argument here http://getkirby.test/docs/reference/objects/toolkit/collection/slice

distantnative:
Screenshot 2025-06-04 at 21 44 28

Should show `int|null $limit = null`

PR #2435, opened 2025-03-20
Add filter feature to reference sections

afbora:
https://github.com/user-attachments/assets/3ef0bf87-b72c-4ff5-986d-426e4946d5c0


bastianallgeier:
That's really nice! I wonder if the filter button is needed at all or if we should always show the input

afbora:
I'm ok with either way. But I think it would be more practical for users to have the filter always visible without having to click any button.

distantnative:
It's my fault Ahmet built the filter button – for me having the search box always visible was a bit waste of space as I'd be more times interested in just seeing more of the actual content of the page and way fewer times inclined to search/filter something, needing that UI element.

distantnative:
I have an idea how to maybe combine both. Let me try something.

bastianallgeier (approved):
So so useful!

PR #2438, opened 2025-04-07
Subscribe meetups calendar

afbora:
Here first step for subscribing to meetups calendar. I tried to add all the information we have to the calendar. Do you think there is anything else we should add?

## Todos

- [x] Adjust to subscribe button in meet page 🙈 
- [x] Use `webcal` protocol to open in an application
- [x] Check the dates and timezones

### DIFF
--- a/site/models/event.php
+++ b/site/models/event.php
@@ -21,6 +21,14 @@ class EventPage extends Page
 		);
 	}
 
+	/**
+	 * Returns the UTC datetime of the event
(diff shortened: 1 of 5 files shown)


afbora:
@bastianallgeier I think now event timezones are correct. I used UTC times for it. I hope that's correct 🤞

bastianallgeier (approved):

PR #2439, opened 2025-04-08
Fix link to plugin page

Sean-OShea:
The developer name and slug has changed. I checked, the three pages still exist with those new URIs


distantnative:
Thanks <3

PR #2440, opened 2025-04-09
Meet events: support end time

distantnative:
### Summary of changes
- Support for `end` time field
- If `end` field is empty, fall back to +2 hours from start time for event duration
- Removed `EventPage:: datetimeUtc()` in favor of `::datetime(timezone: 'UTC')`
- Fixed ical location field: space after comma

### Reasoning
Currently calendars interpret the entries as 30 minutes long.

### DIFF
--- a/site/models/event.php
+++ b/site/models/event.php
@@ -1,32 +1,37 @@
 <?php
 
-use Kirby\Cms\Page;
 use Kirby\Content\Field;
 use Kirby\Toolkit\Str;
 
-class EventPage extends Page
+class EventPage extends DefaultPage
 {
-	public function date(): Field
+	public function date(string $time = 'start'): Field
 	{
 		return parent::date()->value(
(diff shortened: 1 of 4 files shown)


site/models/event.php
```diff
@@ -1,32 +1,37 @@
...
-	{
+	public function datetime(
+		string|null $timezone = null,
+		string $time = 'start',
+	): DateTime {
 		return new DateTime(
-			(string)$this->date(),
-			new DateTimeZone((string)$this->timezone())
+			datetime: $this->date($time),
+			timezone: new DateTimeZone($timezone ?? $this->timezone())
```
afbora: I've tested and this doesn't work correct. Still gives local dates. 
Works with `->setTimezone(new DateTimeZone('UTC'))`.

distantnative: Does it now work for you, @afbora?

afbora (approved):

PR #2443, opened 2025-04-21
Update contrast favicons

lukasbestle:
### Summary of changes

Update the contrast/filled favicons based on our official icon with the inner whitespace removed

### Reasoning

They were previously still based on Kirby’s old icon (different K shape)

### Additional context

We need to make the same change to other repos afterwards; I've saved the Affinity file and the exported variants in our team cloud in "grafiken > favicon".


lukasbestle:
Closing this for now, keeping a reference in the Notion project

PR #2445, opened 2025-05-03
Redesign revenue limit checkbox for more clarity

lukasbestle:
### Summary of changes

- FAQ entries now auto-open when they are being linked to
- The revenue limit info box on the buy page and the help text in the checkout now link to the relevant FAQ entry directly
- Redesigned revenue limit checkout:

**Before:**

Revenue limit checkbox before the change

**After:**

Revenue limit checkbox after the change

### Reasoning

- Added clarity
- Makes the revenue limit less likely to be missed

### DIFF
--- a/site/snippets/faq.php
+++ b/site/snippets/faq.php
@@ -1,10 +1,28 @@
 <div class="faq">
 	<?php foreach ($questions as $question): ?>
-	<details class="details">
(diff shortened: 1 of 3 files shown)


distantnative:
@lukasbestle Sorry this has taken way too long/fell off the table.

I think it's a good change. Only suggestion I would make is to use `mb-3` between the description and the checkbox label:

Screenshot 2025-06-15 at 10 53 44

lukasbestle:
@distantnative Agreed, I have changed it.

PR #2454, opened 2025-05-24
feat: First step llms.txt for docs

afbora:


site/routes/docs.php
```diff
@@ -1,6 +1,32 @@
 <?php
 
+use Kirby\Cms\Page;
+
 return [
+	[
+		'pattern' => 'docs/(:all).md',
```
lukasbestle, line 5: I get that this route avoids having to define a `.md.php` representation template for all individual docs templates, but IMO this route setup is quite confusing and it could look weird to others that we don't use our own content representation feature that's made for exactly this kind of use case.

If we really want to avoid having to create individual representation templates, I think I'd prefer to convert the `llms.md.php` template into a snippet and returning this snippet directly from the route instead of using a virtual page. Makes it simpler and clearer.

content/llms/llms.txt
```diff
@@ -0,0 +1 @@
+Title: LLMS
```
lukasbestle: In contrast, I think for llms.txt it makes sense to go with a route that returns a snippet. Otherwise we would have to create yet another page for llms-full.txt.

afbora:
@lukasbestle Thanks for the review. I've pushed the changes.

bastianallgeier:
I love how practical this approach is. I'm still a bit sad that we can't simply use the content representations, but I guess it would also be a bit tricky to not forget about some of the templates. 

I really wonder about how much we need to post-process the Markdown/Kirbytext though. We have a couple kirbytext parts that we definitely need to resolve. 

- (snippet: ) 
- (reference: )
- (properties: )
- (docs: )
- (api-fields: )
- (field-options: )
- (section-options: )

I think the rest could be understood quite well by an LLM, but those are actively adding content and if we don't apply them, a lot of stuff is simply missing from the docs. 

There could maybe be a global defined at the beginning of the page-content-md snippet. For each field, we still run `$page->text()->kirbytags()` but in our `site/plugins/site/extensions/tags.php` each tag could then render different output based on that global.

afbora:
I think also for:

- (tabs…)(…tabs) - we can remove it from llms.txt
- (columns…)(...columns) - we can remove it from llms.txt
- (image:) - should be like that: `alt text`
- (link:) - should be like that: `[title](https://www.example.com)`
- (glossary:)
- (class:)
- (screencast:)

afbora:
@bastianallgeier Conflict is fixed.

site/plugins/site/composer.json
```diff
@@ -6,17 +6,14 @@
 		}
 	},
 	"require": {
+		"phpdocumentor/reflection-docblock": "^5.3.0",
```
bastianallgeier: `Question`: What's the reason for installing this new package here and removing the phpdoc-parser package? I can't really find where this is used in the llm specifc code.

afbora: Good catch. It's from rebased stuff. I'll fix it.
https://github.com/getkirby/getkirby.com/commit/b90e6f9e638da18e54a8df55b0f4751a5a7f3712

afbora: Done ✅

bastianallgeier:
After going through some of the converted docs, I have to say that I'm not too impressed with the converter. I gave it another shot to create a custom converter which will simply keep the markdown parts but replace the kirbytags: https://github.com/getkirby/getkirby.com/tree/feature/docs-llms-txt-v2

I've also moved to content representations instead of a global route. This way, we can set up individual templates and improve the output of those whenever needed or simply redirect the LLMs to the HTML version if there's no markdown representation. The markdown version becomes more of a enhancement if it's available. But I really think that we have to make sure that the quality of the markdown is good. If this is going to be source for all LLMs to learn about Kirby, it can't be worse than the HTML version.

PR #2466, opened 2025-06-23
feat: Improve playground switch animation

distantnative:
### Todo
- [x] If menu items are clicked too quickly after each other, flickers a bit


site/templates/home.plygrnd.php
```diff
@@ -0,0 +1,4 @@
+<?php snippet('templates/home/playground/filesystem') ?>
+<?php snippet('templates/home/playground/backend') ?>
+<?php snippet('templates/home/playground/medium') ?>
+<?php snippet('templates/home/playground/audience') ?>
```
distantnative: `Question:` Is this overkill?

site/snippets/templates/home/playground/header.php
```diff
@@ -14,42 +14,54 @@
...
-					<?php endforeach ?>
-					<li><a class="font-bold more" href="/love">Your ideas &rarr;</a></li>
-				</ul>
+				<div class="sticky" style="--top: var(--spacing-2)">
+					<ul class="font-mono text-sm pt-6 mb-6">
+						<?php foreach ($stories as $option): ?>
+						<li>
+							<a
+								<?php e($story === $option, 'aria-current="true"') ?>
+								href="/home.plygrnd?your=<?= $option->slug() ?>"
```
distantnative: `Comment:` Don't think we can call a content representation of the home page differently.

assets/js/layouts/playground.js
```diff
@@ -1,69 +1,133 @@
...
+		}
+	}
+	get figure() {
+		return this.$el.querySelector(".playground-header-figure");
 	}
 
 	async loadHtml(link) {
-		const response = await fetch(link);
+		const response = await fetch(link, {
+			cache: "no-store",
```
lukasbestle: `Question`: Why do we need to disable the browser cache?

assets/js/layouts/playground.js
```diff
@@ -1,69 +1,133 @@
...
+				setTimeout(() => {
+					oldImage.remove();
+				}, 700);
 
-		new Image().src = newUrl;
+				resolve();
+			};
+
+			// append new image behind current image
+			this.wrapper.appendChild(newImage);
```
lukasbestle, line 72: `Finding`: I think the flickering on quick theme or story changes happens because multiple new images get loaded into the DOM before the first old one is removed.

`Idea` how we could solve this:
- When a `newImage` is appended to the DOM, set a global flag like `pendingUpdate`.
- Before executing the logic in `replaceImage()`, check if the flag is set. If so, set the new target to another global prop like `finalTarget` and return immediately.
- Before resolving the top-level promise, unset the `pendingUpdate` prop. Also check if something is in the `finalTarget` prop. If so, unset it and call `replaceImage(finalTarget)` to start a second update cycle for the final target.

Maybe it could be done even cleaner with a throttle helper, but I think this should already work quite well.

lukasbestle (commented):
`General comment`: I love the crossfade transition for the theme change. Not sure about using it for the story change as well though. There I think the fade to white was better at "setting the stage" for the new story. Also the old loading order ensured that the image and playground are replaced at the same moment. Now these two operations happen in parallel and potentially out of sync, which can lead to visual jumping.

`Suggestion`: We could have two modes. One for the theme switcher that just replaces the image with the new logic that automatically crossfades. And the old mode for switching stories.

assets/js/layouts/playground.js
distantnative: pinging @bastianallgeier - that's from his changes

site/templates/home.plygrnd.php
lukasbestle: If I understand it correctly, this is to make the dynamically loaded HTML slimmer and omit all the header and footer code, right?

I think it's not overkill. For that, it would have to have a significant downside that wouldn't make it worth it. But this is quite a compact file and the only thing we would need to maintain is the list of snippets to include. But that's fine IMO.

`Suggestion`: We could name it `home.partial.php` to make the name less cryptic.

assets/js/layouts/playground.js
lukasbestle: Could have been necessary because of the `Expires` header during the current sale, see my post in Discord #docs today. But we shouldn‘t need to disable the cache indefinitely. Once we pushed the new playground code, that will again be cached properly after the sale is over.

bastianallgeier: Yes, it was completely impossible to work on the switch without disabling it and once I saw the caching issues I got really scared that they could have an effect in the final version as well. I agree that it would be better to have caching enabled.

distantnative: I removed it

assets/js/layouts/playground.js
distantnative: @lukasbestle I added a variant as you suggested. I think this "pending queue" works quite well (for the already edge case).

assets/js/layouts/playground.js
```diff
@@ -127,6 +136,12 @@ export class Playground {
 		// and stop loader animation
 		this.figure.dataset.theme = this.theme;
 		this.figure.classList.remove("loading");
+		this.isUpdating = false;
+
+		if (this.pendingUpdates.length > 0) {
+			const { target, link } = this.pendingUpdates.shift();
+			this.switchTo(target, link);
```
lukasbestle: `Question`: Do we need to execute all of them in order? Or phrased differently, does this actually need to be a queue? If the user clicks multiple things, I think it would be completely fine (and even expected) that everything in between is dropped and only the final requested state is loaded. This would mean that we can turn `this.pendingUpdates` into a simple object that just gets replaced each time.

lukasbestle (commented):
Works really great for me in Safari, Firefox and Chrome. 🎉

assets/js/layouts/playground.js
distantnative: @lukasbestle Good idea. Changed it. Also moved the current link change to when image is shown and content is changed. Felt better on slow connections that all 3 update at the same time. Even if then initially only the loader starts spinning.

lukasbestle (approved):
Absolute bliss 😍

PR #2474, opened 2025-07-09
[v4] Search feature for arcihived docs with Loupe SQLite local search engine

### LINKED ISSUE #2471: Algolia search is broken for archived docs
https://v4.getkirby.com

afbora:
### Summary of changes

- Added `Loupe` SQLite search engine - Implemented a new local search solution that doesn't require external services
- Created content indexing system - Built content index functionality to support the new search engine
- Removed `Algolia` search dependency - Eliminated the external Algolia search service and related code
- Fixed asset path checking - Updated to use str_starts_with() for more efficient asset path validation
- Removed `Kosmos` and `Plugin` areas - Cleaned up related search areas that is no longer needed

Related issue: #2471 

image


bastianallgeier:
This is awesome. I'm a bit worried though that we cannot simply switch to this solution for later archived versions. There needs to be some kind of A/B state where we can switch between Loupe or Algolia based on the "archived" option in the config. When we replace the entire search plugin, this won't work. Otherwise we would need to repeate the same steps with every major release launch. Maybe we can split this PR into multiple steps to make it easier? 

We could for example have `site/config/search.algolia.php` & `site/config/search.loupe.php` and also two plugins for algolia and for loupe and either one of them gets loaded. Do you think this is feasible?

afbora:
Thank you @bastianallgeier As I understand it, we want to provide both search engines and if it's archived docs, it should search on loupe, otherwise it should search on algolia. Did I understand correctly?

afbora:
I've created new PR: https://github.com/getkirby/getkirby.com/pull/2475

distantnative:
@afbora Shall we close this in favor of your new PR?

PR #2475, opened 2025-07-12
[v4] Add Loupe sqlite based search engine for archived docs

afbora:
### Summary of changes

- Added `Loupe` SQLite search engine 
- Created whole content index (site/config/loupe/loupe.db)
- Removed `Kosmos` and `Plugin` areas
- Hidden algolia logo in search dialog when current docs are archived
- Added CLI script for Kirby Loupe re-indexing
- Adjusted search controller to use correct search engine (Archived docs: Loupe, Fallback: Algolia)
- Fixed asset path checking (there was an error about search results)

Related issue: #2471 
Related PR:#2474

image


site/config/search.php
```diff
@@ -82,12 +82,128 @@
 			'release-35'
 		]
 	],
+	'loupe' => [
```
distantnative: This feels a bit weird as we have on the top level all config for Algolia but then this one key that defines all options for Loupe, some of them very similar to other top-level options. I would suggest to create separate files for these, `config/search.algolia.php` and `config/search.loupe.php`.

site/controllers/search.php
```diff
@@ -3,22 +3,35 @@
...
+	$area = trim(get('area', ''));
 	$results = null;
 
-	if (empty($query) === false) {
-		$params = [
-			'hitsPerPage'           => (int)get('limit', 50),
-			'attributesToHighlight' => false,
-			'attributesToSnippet'   => '*'
-		];
+	if ($query !== '') {
```
distantnative, line 10: `Suggestion:` To make this less nested and easier to read, I am wondering if it would be better to work with an early return, e.g. to check for `if ($query === '') { return [...]  }`.

site/plugins/cdn/index.php
```diff
@@ -55,7 +55,7 @@
 		'url' => function (App $kirby, $path, $options): string {
 			static $original;
 
-			if (preg_match('!assets\/!', $path ?? '')) {
+			if (str_starts_with($path ?? '', 'assets')) {
```
distantnative, line 58: `Question:` Why is this change necessary/how does it relate to this PR?

distantnative (commented):
What is the reason for creating an own Loupe plugin instead using an existing one like https://plugins.getkirby.com/arnoson/loupe? Genuinely asking, you probably have good reasons, I would just like to understand it and I don't have any knowledge about them.

I would suggest to rename `search` to `search-algolia` then. 

And currently we have a lot of repeated code/classes between the two plugins. Would be great to look into whether we can make them a bit more DRY. Share classes or code between them.

afbora:
> And currently we have a lot of repeated code/classes between the two plugins. Would be great to look into whether we can make them a bit more DRY. Share classes or code between them.

@distantnative Yes, I considered that, but there are a few reasons why I made it standalone instead of shared:

- We might use something else instead of Loupe in the future.
- The Loupe plugin will never work (be active) on our current site.
- It should be easy to move between branches.

site/plugins/cdn/index.php
afbora: Throws an exception when search result contains `assets` string.

```
Whoops\Exception\ErrorException thrown with message "file_exists(): Passing null to parameter #1 ($filename) of type string is deprecated"

Stacktrace:
#15 Whoops\Exception\ErrorException in \getkirby\site\plugins\cdn\src\Optimizer.php:55
#14 file_exists in \getkirby\site\plugins\cdn\src\Optimizer.php:55
#13 Kirby\Cdn\Optimizer:cachebust in \getkirby\site\plugins\cdn\index.php:64
#12 Kirby\Filesystem\F:{closure} in \getkirby\kirby\src\Cms\Url.php:84
#11 Kirby\Cms\Url:to in \getkirby\kirby\config\helpers.php:627
#10 url in \getkirby\site\templates\search.php:77
#9 include in \getkirby\kirby\src\Filesystem\F.php:425
#8 Kirby\Filesystem\F:loadIsolated in \getkirby\kirby\src\Filesystem\F.php:364
#7 Kirby\Filesystem\F:Kirby\Filesystem\{closure} in \getkirby\kirby\src\Filesystem\F.php:372
#6 Kirby\Filesystem\F:load in \getkirby\kirby\src\Toolkit\Tpl.php:36
#5 Kirby\Toolkit\Tpl:load in \getkirby\kirby\src\Template\Template.php:163
#4 Kirby\Template\Template:render in \getkirby\kirby\src\Cms\Page.php:1021
#3 Kirby\Cms\Page:render in \getkirby\kirby\src\Cms\App.php:794
#2 Kirby\Cms\App:io in \getkirby\kirby\src\Cms\App.php:1201
#1 Kirby\Cms\App:render in \getkirby\index.php:7
#0 require in \Herd\resources\app.asar.unpacked\resources\valet\server.php:139

````

So I changed to starts with check not contains. Is this more accurate? Not only this PR.

**Comment by distantnative** [comment]
@afbora @bastianallgeier I have merged both plugins into one plugin which allows us better to reuse shared code. It seems to work fine, but I had to increase the memory limit to be able to run the script which could be an issue on the server.

We should also not only merge this into the `archive/4.0` branch but add it to the repository in general, right?

File: scripts/loupe-index [hunk 4]
```diff
@@ -0,0 +1,41 @@
...
+
+use Kirby\Cms\App;
+use Kirby\Search\LoupeSearch;
+
+// check if we are indeed on the command line
+if (php_sapi_name() !== 'cli') {
+    die();
+}
+
+ini_set('memory_limit','2048M');
```
distantnative: Without this, running the script locally would end up in a memory limit exceeded error. That could cause problems on the server.

distantnative: Ok I was able to remove it locally for me when refactoring the search index to yield results and add them in batches to the loupe database.

afbora:
Thanks @distantnative for taking care. Did you tested via non-archived env? I'm getting some errors. You'll see when you set `archived = false` from config file.

site/controllers/search.php
```diff
@@ -3,29 +3,24 @@
...
+		$results = search(
+			area:  $area,
+			query: $query,
+			page:  param('page', 1),
+			limit: get('limit', 50)
+		);
 	}
 
 	return [
 		'results'    => $results ?? [],
```
afbora, line 20:
```suggestion
		'results'    => $results ??= [],
```

We should use like that or we need to define `$results` variable as `null` as we did before.

distantnative: Why? Where else are we passing `$results`?

distantnative:
@afbora Not on this branch. As it is based on the archived branch where a lot of content has been removed, it will directly throw errors when changing the config as those pages cannot be found (and then e.g. calling `->menuUrl()` on null). We would first have to set up a new branch for the live site and copy over the plugin.

site/controllers/search.php
afbora: We are using `$result` variable in next line.

distantnative: But we also check with `?` if it exists. If we assign `[]` that would break as we cannot call a method on an array.

afbora: AFAIK `?->` is for `null` check but `$results` is not defined if query is empty.

afbora:
> As it is based on the archived branch where a lot of content has been removed, it will directly throw errors when changing the config as those pages cannot be found (and then e.g. calling ->menuUrl() on null). We would first have to set up a new branch for the live site and copy over the plugin.

OK, fine. It is not important for Algolia to work correctly in archived documents, but both must work correctly in current documents.

site/controllers/search.php
distantnative: Ah you are right. I thought this would work also for non-existing. I think it needs to be `$results?->pagination() ?? null` then it seems to work.

distantnative:
@afbora Fully agreed - but we cannot test that with this branch. I have created a new branch based on `main`: https://github.com/getkirby/getkirby.com/pull/2487

distantnative:
Closing this in favor of https://github.com/getkirby/getkirby.com/pull/2487 to start from our `main` branch (and then we can port it to the `archive` branch too.

PR #2479, opened 2025-07-23
Update guide.txt

reidarpritzel:
Missing an s in "controllers" in the code example path. 

Previously:
site/controller/site.php

Correct:
site/controllers/site.php

## Description
Missing s in "controllers" in code example path

### Summary of changes
Added an s

### Additional context
This is a very minor thing, but could be misleading for some.


bastianallgeier (approved):

bastianallgeier:
thanks for the fix!

PR #2484, opened 2025-08-03
Dead link in reference-article.txt

Sean-OShea:
Hopefully the link i provided is the right one once again!

## Description
The current link points to a page that doesn't exist. I checked the archived docs but couldn't find this page, so i'm not sure if i missed something. 

### Summary of changes
Modified the link to a page that has the expected content


distantnative:
Thank you!

PR #2485, opened 2025-08-04
fix: syntax error

pReya:
## Description
- add missing "closing code segment" syntax


bastianallgeier (approved):

bastianallgeier:
Thanks for the fix @pReya

PR #2486, opened 2025-08-07
feat: llms.txt and markdown version for all doc pages

bastianallgeier:
## Description

This PR builds on @afbora's initial draft PR https://github.com/getkirby/getkirby.com/pull/2454 and implements markdown versions for all doc and reference pages, as well as a new /llms.txt overview page.


site/models/reference-classes.php
```diff
@@ -69,6 +69,11 @@ public function children(): Pages
 		return $this->children = Pages::factory($children, $this);
 	}
 
+	public function childrenUnfiltered(): Pages
```
distantnative: `Suggestion`: I would go for another name as the main method doesn't really filter anything per se - so another name might make it easier to understand what's the purpose of this one is.

site/plugins/site/extensions/fieldMethods.php
```diff
@@ -1,11 +1,19 @@
...
+use Kirby\Text\KirbyTags;
 use Kirby\Toolkit\Collection;
 use Kirby\Toolkit\Obj;
 use Kirby\Toolkit\Str;
 
 return [
+	'convertToMarkdown' => function ($field) {
+		return $field->value(function ($value)	{
+			return kirbytagsToMarkdown($value);
+		});
```
distantnative: `Nitpick`

```suggestion
		return $field->value(fn ($value) => kirbytagsToMarkdown($value));
```

site/plugins/site/helpers/markdown.php
```diff
@@ -0,0 +1,132 @@
...
+{
+	$text ??= '';
+
+	// replace three consecutive newlines with two
+	$text = preg_replace('/\n\n\n/', "\n\n", $text);
+
+	return trim($text);
+}
+
+function kirbytagsToMarkdown(string|null $text): string
```
distantnative, line 19: `Question:` Couldn't we handle this in our Marsdown class? Similarly to how it's done in the individual Kirbytags with checking the option?

site/plugins/site/src/Reference/Reflectable/ReflectableClassMethod.php
```diff
@@ -99,6 +99,23 @@ public function isUnstable(): bool
...
+		}
+
+		$object = '$' . strtolower($this->class(true));
+
+		if ($returns->isImmutable() === true) {
+			return 'This method does not modify the existing `' . $object . '` object but returns a new object with the changes applied. (link: docs/guide/templates/php-api#immutable-objects text: Learn more →)';
+		}
+
+		return 'This method modifies the existing `' . $object . '` object it is applied to and returns it again.';
+	}
```
distantnative, line 117: `Suggestion`: I would think this belongs into the `Returns` class.

site/models/reference-classes.php
bastianallgeier: Good catch. I didn't end up using the method at all, but forgot it in the model.

site/plugins/site/helpers/markdown.php
bastianallgeier: I'm not sure I can follow here.

site/plugins/site/src/Reference/Reflectable/ReflectableClassMethod.php
bastianallgeier: I thought so too at first, but the Returns class does not have access to the parent class.

site/plugins/site/helpers/markdown.php
distantnative: Our Marsdown class already does the parsing for columns, tabs etc. - maybe even more reliably (maintained) in the long term to not add another feature that applies its own regex to handle the same things, but rather in Marsdown decide (like with the single kirbytags) "if the kirbytags option is false, return this markdown, otherwise return this HTML".

bastianallgeier: Ah, now I got you. The problem is that we don't want the Markdown to be parsed at all. That's why the Marsedown parser never runs for those .md representations. Otherwise, your suggestion would definitely be better.

distantnative (approved):

afbora:
Are we planning llms-full.txt?

bastianallgeier:
@afbora I don't have enough knowledge to say if it is needed or not.

PR #2487, opened 2025-08-14
feat: Loupe search for archived docs

distantnative:
I experienced memory allocation exceeded errors again when I tried to create the index. If we cannot fix this reliably, I am not sure if Loupe will work for us.


afbora:
Getting error for me

image

distantnative:
Pinging @lukasbestle as this is about the security page model.

For me I run again and again in memory allocation errors. It seems very memory hungry, which could be a problem on the server.

lukasbestle:
@afbora The error in `SecurityPage::kirbyVersionNoVulnerabilities()` can only occur if `$securityPage->incidents()` is empty. Those are retrieved from the GitHub API. Please try clearing the cache and check if you have the GitHub API key in `site/config/keys.php`.

afbora:
Thanks @lukasbestle Yes, after setting GitHub key, I can able to index entire site successfully.

image

distantnative:
I realized combining the script and having that called by the GitHub CI doesn't make sense. It works for Algolia cause the index we sent from the GitHub action will be the same as later on the server. But the Loupe database actually needs to end up on the server, not temporarily in the CI action.

afbora:
Yes, the Loupe DB will be a static database created only once. We can create it locally and push it. There's no need to run GitHub CI for archived documents.

distantnative:
I have split the scripts again so that the CI action only runs the Algolia index and the Lupe index can be run manually on the archived version branches to commit the Loupe index database.

I have also added a line to increase the memory limit for the Loupe script as that ran always into errors for me. With the increased memory limit it does work for me.

PR #2489, opened 2025-08-21
Improvements and updates to plugin guide

lukasbestle:
### Summary of changes

- Better and more detailed explanations on plugin naming
- Updates of outdated info
- Various minor fixes

### Reasoning

As discussed in #headquarter

### DIFF
--- a/content/docs/1_guide/17_plugins/1_custom-plugins/guide.txt
+++ b/content/docs/1_guide/17_plugins/1_custom-plugins/guide.txt
@@ -58,11 +58,15 @@ It is also used to identify your plugin during the update check. If you publish 
 
 ### Named arguments
 
+As soon as your plugin gets more complex and may need to provide additional information (see below), it can make sense to use PHP's named arguments to make your plugin registration more clear:
+
 ```php "/site/plugins/your-plugin/index.php"
 Kirby::plugin(
 	name: 'my-name/hello-world',
 	extends: [
-		…
+		'snippets'  => […],
+		'templates' => […],
+		'hooks'     => […]
 	]
 );
 ```
--- a/content/docs/1_guide/17_plugins/1_custom-plugins/guide.txt
+++ b/content/docs/1_guide/17_plugins/1_custom-plugins/guide.txt
@@ -67,7 +67,7 @@ To summarize:
 | - | -------------------------- | ------------------------ |
 | ✅ | `my-name/my-plugin` | `my-name/kirby-my-plugin` |
 | ❌ | `my-name/kirby-my-plugin` | `my-name/kirby-my-plugin` |
-| ❌ | `my-name/kirby5-my-plugin` | `my-name/kirby5-my-plugin` |
+| ❌ | `my-name/kirby(kirby-version: major)-my-plugin` | `my-name/kirby(kirby-version: major)-my-plugin` |
 </info>
 
 ### Named arguments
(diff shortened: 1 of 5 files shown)


content/docs/1_guide/17_plugins/1_custom-plugins/guide.txt
```diff
@@ -166,7 +192,7 @@ This is great if you want to point to your own license document.
 
 ```php
 App::plugin(
-  name: 'my/plugin',
+  name: 'my-name/hello-world',
```
distantnative: `Suggestion:` `hello-world` might a bit random here. How about `my-name/my-plugin`?

content/docs/1_guide/17_plugins/3_plugin-setup-basic/guide.txt
```diff
@@ -141,5 +141,4 @@ Once your plugin is ready, you can publish it with the following steps:
 1. Push your code to your Git repository.
 2. Create a Git tag with the version number of your release.
 3. Publish your plugin on [Packagist](https://packagist.org) to enable installation via Composer (only needed once, Packagist will pick up new Git tags automatically after that).
-4. If you want to advertise your plugin in the [plugin directory](https://plugin.getkirby.com), send us an email to <support@getkirby.com> with a link to your Git repository and a cover (cover image in 2:1 format or a short code snippet).
-5. Tell everybody about it on the (link: https://forum.getkirby.com text: forum) in the plugins category.
+4. (link: docs/guide/plugins/best-practices#promote-your-plugin text: Promote your plugin in our plugin directory and the forum)
```
distantnative, line 144: I think it would still make sense here to let people know to reach out to us via email if they want their plugin listed (as there currently is not yet a way to add these themselves).

content/docs/1_guide/17_plugins/7_best-practices/guide.txt
```diff
@@ -105,10 +107,14 @@ Add a good description to your plugin repo so that people looking for a specific
 
 On GitHub, you can tag your plugins. Don't forget to add the following tags in addition to those that describe the purpose of your plugin:
 
-- kirby3
+- kirby5
```
distantnative: `Suggestion:` Maybe worth a small KirbyTag for our site plugin that outputs the current version number? Then this would stay up to date automatically.

content/docs/1_guide/17_plugins/3_plugin-setup-basic/guide.txt
distantnative: I see below it's still in the best practices. But since you mention promoting it in the directory here too, I would mention it here as well.

lukasbestle: We had that email link before, but it was duplicate content as we had the same info both on the plugin setup page (here) and in the best practices. We had actually forgotten to update one of them before, so I think it makes sense to finally consolidate this in one location and provide cross-links to the canonical location. The section in the best practices does a better job at explaining than this did IMO.

distantnative: I think we need a direct link to the best practices then. I don't know we can assume here that the user will definitely get to the best practices page as well. And I agree with your general thought but since we do actually talk about the topic here, I think we also need to touch up on this here too. Alternatively, we should not touch the topic here at all and only on the best practice page.

lukasbestle: But that direct link to best practices is exactly what you are commenting on? :D

distantnative: 🙈 I guess it's time to call it a week 🤣

distantnative (approved):
👍 maybe @texnixe wants to take a look as well

lukasbestle:
Merging this for now. We can still improve and update it later.

PR #2492, opened 2025-09-08
Update guide.txt

thejohnnyhill:
Updating to replace `k-writer` (deprecated as of 5.0.0) with `k-writer-input`

## Description
I've been building some custom components off of this (very handy) guide, but ran into an issue with some of the live editable previews. I dug into the Vue component documentation for Kirby, and found out that the `k-writer` element is deprecated and no longer works in Kirby 5.0.0, in favor of `k-writer-input`. Updated this page to reflect that change. 

Might also recommend adding a specific callout about this issue for users who might be on an older version of Kirby.


distantnative (approved):

distantnative:
Thank you for catching this and providing the fix yourself 🙏

PR #2493, opened 2025-09-10
Update cookbook-recipe.txt regarding HTML/listitems

marcus-herrmann:
## Description
I fixed the code example for the linking to children and logout part, regarding proper list HTML.

### Summary of changes
Added `<ul>`, `<li>` where needed.

### Reasoning
HTML: Code example showed `<li>`s as direct children of `<nav>`, which is against the spec.


bastianallgeier:
Good catch! I think the original intention was to not use any li elements at all. But I both ways work as long as it is consistent.

PR #2496, opened 2025-10-11
Fix wrong centering on mobile of "Made in Europe" footer

eddiedale:
This seems unnecessary on desktop also as far as I can tell, but more importantly causes unwanted horisontal centering (at least seems unwanted in my eyes) with the changing to flex column on @media screen and (max-width: 50rem)

Better just removed, and left aligned all the way?

eddie-11-10-2025-13 56 14


lukasbestle:
Thank you :)

PR #2498, opened 2025-10-16
Update guide.txt

AndreKelling:
fix forum plugins link

## Description

i guess the link to forum with plugins topic was getting broken in the past.

### Summary of changes

just fix the link

### Reasoning

-

### Additional context

-


bastianallgeier (approved):
Thanks for the fix

PR #2504, opened 2025-10-25
refact: Layout changes for Buy page

distantnative:
## Description

A first idea. Trying to move in smaller steps. This could be a first one.
Screenshot 2025-11-02 at 10 51 39

### DIFF
--- a/assets/css/site/utilities.css
+++ b/assets/css/site/utilities.css
@@ -96,6 +96,9 @@ img {
 .max-w-xl {
 	max-width: 36rem;
 }
+.max-w-2xl {
+	max-width: 48rem;
+}
 
 .rounded-xl {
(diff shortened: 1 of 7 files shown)


lukasbestle (commented):
I like it a lot! It puts the information much more front-and-center. I also love the rewording of the revenue limit FAQ and headline, much clearer.

Design-wise I'm not sure about the "Which license type do I need?" link, but I don't have an alternative suggestion. Maybe something for @bastianallgeier.

Then there is a minor issue with the alignment on larger screens. If the text doesn't have the same number of lines, the checklist is not on the same level, making the visual comparison harder. Maybe we could use CSS Grid here to align these?

Image

bastianallgeier (approved):

PR #2506, opened 2025-11-02
refact: Integrate good cause list into Buy page

distantnative:
## Description

- Integrates the good cause details directly in the Buy page instead of a separate subpage. The old subpage redirects.
- Shows more good cause examples (in a separate step, we might want to add more/update the good cause cases).

Screenshot 2025-11-02 at 11 33 36


site/snippets/templates/buy/good-cause.php
```diff
@@ -6,40 +6,48 @@
...
+							<span class="block shadow-2xl mb-3" style="--aspect-ratio: 3/4">
+								<?= img($image = $case->image(), [
+									'alt' => $image->alt()->or('Screenshot of the ' . $case->title() . ' website'),
+									'src' => [
+										'width' => 300
+									],
+									'srcset' => [
+										400,
+										800,
+									]
```
distantnative, line 41: I realized these probably need to be updated now that we are having more columns.

bastianallgeier (approved):

site/snippets/templates/buy/good-cause.php
bastianallgeier: It could be slightly smaller. But I think it's not super important. When you resize the window we still have a version where the screenshots are roughly 400 pixels wide.

PR #2508, opened 2025-11-04
Keyboard accessibility

medienbaecker:
## Description

I made some (hopefully) low impact changes to make sure interactive elements are accessible to keyboard users on [getkirby.com](https://getkirby.com/).

### Summary of changes

1. The copy buttons in code blocks had their outline disabled and were only accessible through "hover". Now the outline is enabled and it can be accessed through `focus-within`, too.
2. `summary` elements had their outline (globally) disabled, making it impossible for keyboard users to see where the focus is. I've removed that in two different places. ~~One is for the filesystem component, where you use "decorative" `details`/`summary` elements. I'm not sure if that's a good idea anyways, but I've added a negative `tabindex` instead of the hidden outline there. Now the home page doesn't have this focus black hole anymore.~~
I've just noticed you use this filesystem snippet [in the docs](https://getkirby.com/docs/guide/tour#where-the-content-lives), too. I didn't know the folders were collapsible, nice! So I've reverted adding the negative `tabindex` for now.

getkirby com_buy

3. When a "topbar" is active ("🎉 Kirby 5 is here!"), "skip to content" was not the first element. I've swapped their positions in the DOM, making it the first focusable element again.
4. The outlines of your dark buttons on light backgrounds were hard to see, so I've added an `outline-offset`:

getkirby com_partners


bastianallgeier (approved):
Sorry for the delay. The changes are great and all work as expected. Thanks for your help 💛

PR #2510, opened 2025-11-08
Exclude 'vendor' directory from scanning

kllmanu:
to exclude directories one has to use `@source not`


bastianallgeier:
Duplicate: https://github.com/getkirby/getkirby.com/pull/2497

PR #2512, opened 2025-11-08
Update cookbook-recipe.txt with Git repo info

kllmanu:
I feel like this info is required, because many people don't use git for their kirby projects (me being one of them 😜). If you don't have a git repository initialized, Tailwind won't find any source files to collect classes from. It would be nice if we would tell this info to people following this guide.


bastianallgeier:
It looks like a gitignore file could be enough https://github.com/tailwindlabs/tailwindcss/discussions/18405 I'm not a Tailwind expert though.

PR #2517, opened 2026-01-04
docs: add hint at 'method' option in API routes

tillsanders:
## Description

I recently tried to add a custom endpoint for a plugin expecting to receive POST requests. Without the `method` option set to 'POST', Kirby's router will only return 404. It looks like this option is undocumented, so I'm at least adding a little hint in the code example.

### Summary of changes

Added optional `method` option to code sample in docs on custom api endpoints.


tillsanders:
Not sure I should have forked from main. Should I fork from develop and open a PR to merge there, too? Happy to open another PR using the correct branches!

distantnative:
@tillsanders Looks like you already branched off/forked from main - just point this PR back to main instead of develop please. We only use develop for the website sometimes when staging larger changes, so I guess it's currently quite behind.

bastianallgeier (approved):

PR #2518, opened 2026-01-05
code template correction : aria-current value on breadcrumb

Sean-OShea:
For the current page, the aria-current must be "page" and not "location"

Ref : https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-current#values


afbora (approved):

afbora:
Thanks 👍

PR #2520, opened 2026-01-06
Update partner profile Dominik Laube

dominiklaube:


bastianallgeier:
I think I would actually keep the larger image to be honest. The new one is quite small.

dominiklaube:
@bastianallgeier thanks for your hint. I updated the image quality.

PR #2521, opened 2026-01-09
Image performance improvements

lukasbestle:
### Summary of changes

- Only use lazy-loading for below-the-fold images
- `srcset` for full width KirbyText images

### Reasoning

- Increase loading performance
- Improve match of image sizes to screen (e.g. small screens, Retina desktop screens)

### DIFF
--- a/site/plugins/site/extensions/tags.php
+++ b/site/plugins/site/extensions/tags.php
@@ -300,7 +300,8 @@ $tags['screencast'] = [
 			'url'    => $tag->value,
 			'poster' => $tag->poster ? $tag->file($tag->poster) : $tag->file('youtube.jpg'),
 			'title'  => $tag->title ?? null,
-			'text'   => $tag->text ?? null
+			'text'   => $tag->text ?? null,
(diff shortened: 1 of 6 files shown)


site/plugins/site/extensions/tags.php
```diff
@@ -300,7 +300,8 @@
 			'url'    => $tag->value,
 			'poster' => $tag->poster ? $tag->file($tag->poster) : $tag->file('youtube.jpg'),
 			'title'  => $tag->title ?? null,
-			'text'   => $tag->text ?? null
+			'text'   => $tag->text ?? null,
+			'lazy'   => $tag->data['field']->key() !== 'screencast'
```
distantnative, line 304: @lukasbestle Trying to understand this line: You want it to be lazy always if not inside a field called `screencast`?

lukasbestle: Yes. The `screencast` field is always rendered at the top of the page, so the image shouldn't be loaded lazily. For all other uses of the tag I assume they are rendered further down the page and can therefore be loaded lazily.

distantnative (approved):

PR #2522, opened 2026-01-11
Add model for some backend validation

texnixe:


bastianallgeier (commented):
It all looks good, except the business type validation. I feel like it could be a bit too strong.

lukasbestle:
Also thought that first, but it only checks if the first character is something weird. The rest of the string can contain whatever character.

bastianallgeier:
Oh, right. Reading regex. Still not easy 😆. Then I think it's fine. Might be even too weak. But I would say it's a good first improvement.

bastianallgeier (approved):

PR #2523, opened 2026-01-11
Update virtual pages guides

texnixe:
## Description
This PR removes the advanced section from the "Content from a database guide" and adds a link to the WIP database-storage plugin instead. We can later add a cookbook recipe that explains the concept in detail.

The second part is an update of the mixed content guide based on the MixedStorage class from the forum, plus some additions.

### DIFF
--- a/content/docs/1_guide/19_virtual-content/3_content-from-database/guide.txt
+++ b/content/docs/1_guide/19_virtual-content/3_content-from-database/guide.txt
@@ -2,41 +2,52 @@ Title: Content from a database
 
 ----
 
-Description: Create pages from a database and edit, add or delete entries in the Panel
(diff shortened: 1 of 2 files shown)


distantnative (commented):
I think this is the right change for now - only thing that jumped at me was whether we really want to call it "work-in-progress plugin" and not something like "our database storage plugin (currently in alpha)"?

bastianallgeier:
@distantnative I would prefer this as well. It's actually quite stable already. It has missing support for some actions, but overall we could totally put it in beta as well.

texnixe:
Ok, fixed the wording, going to merge now, we can update it again once the plugin status has changed.

PR #2533, opened 2026-03-09
Feature/kosmos integration

texnixe:
## Description
This PR adds support for merging future "virtual" kosmos episodes from kosmos.getkirby.com with the current file based episodes.

This allows us to have a single source of truth for all newsletters and to not have to create extra versions to be published on the website.

Can be tested locally using this branch together with the kosmos repo.


bastianallgeier:
Is there anything else that I need to configure locally. http://kosmos.test/kosmos.json is empty in my case.

site/models/kosmos.php
```diff
@@ -0,0 +1,88 @@
...
+	{
+		if ($this->children instanceof Pages) {
+			return $this->children;
+		}
+		
+		$data = [];
+		$issues = [];
+		
+		try {
+			$request = Remote::get('http://kosmos.test/kosmos.json');
```
lukasbestle: `Todo`: Please move this hostname to the config.

site/models/episode.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+use Kirby\Cms\Page;
+
+class EpisodePage extends Page
```
lukasbestle: `Nitpick`: For consistency with `kosmos-issue` (we mostly use the area prefixes for getkirby.com templates), I'd vote for `kosmos-issue-virtual` or similar. If we name it `episode`, the connection to Kosmos and the implementation difference between `kosmos-issue` and `episode` would get lost.

site/templates/episode.php
```diff
@@ -0,0 +1,83 @@
...
+<?php slot('sidebar') ?>
+<style>
+    .with-sidebar {
+        grid-gap: 9rem;
+    }
+</style>
+
+<nav>
+    <div class="mb-12">
+        <p class="h1 color-gray-400 mb-1"><a href="/kosmos">Kosmos</a></p>
```
lukasbestle: `Suggestion`: Can we DRY the parts that are shared with the existing `kosmos-issue` template? Could be via snippets or maybe even a shared template that uses a conditional in the part that differs.

site/templates/episode.php
```diff
@@ -0,0 +1,83 @@
+<?php
+layout('article') ?>
```
lukasbestle: `Nitpick`: For consistency with other templates, these PHP one-liners should be in a single line including the open and close tags:

```suggestion
<?php layout('article') ?>
```

(same in some other places as well)

site/models/kosmos.php
```diff
@@ -0,0 +1,88 @@
...
+			);
+			
+		} catch (Exception) {}
+		
+		return $this->children = $this->subpages()->add(Pages::factory($issues, $this));
+	}
+	
+	/**
+	 * Creates a collection of `VirtualFile` objects from array data that comes
+	 * from the partners API's `apiArray()` file method
```
lukasbestle:
```suggestion
	 * from the Kosmos API's `apiArray()` file method
```

Speaking of which, can we DRY this method?

site/snippets/toc.php
```diff
@@ -1,5 +1,5 @@
 <?php
-$items        = $page->text()->toToc($tag ?? 'h2');
+$items        = $items ?? $page->text()->toToc($tag ?? 'h2');
```
lukasbestle:
```suggestion
$items      ??= $page->text()->toToc($tag ?? 'h2');
```

site/templates/episode.php
```diff
@@ -0,0 +1,83 @@
...
+
+<?php slot('sidebar') ?>
+<style>
+    .with-sidebar {
```
lukasbestle: `Nitpick`: The template and some snippets use spaces for indentation. Could you please consistently convert to tabs?

texnixe:
@bastianallgeier kosmos.json is only filled if there are kosmos editions with the publish toggle set to `true`. I added this to prevent premature publishing of newsletters to the website.

site/templates/episode.php
texnixe: I tried hard, but somehow couldn't get it to work... Edit: got it to work.

site/templates/episode.php
texnixe: No idea how this happened, reformatting the file must have introduced this mess. I wonder what wrong settings I have in my IDE.

lukasbestle: Maybe we should all jump on Mago to have a Prettier-like "consistency-first and preference-second" formatter. 😄

site/models/kosmos.php
texnixe: Should we, given that one has content and the other doesn't?

site/templates/kosmos-issue-virtual.php
```diff
@@ -0,0 +1,30 @@
...
+
+<?php snippet('kosmos/sidebar') ?>
+<?php snippet('kosmos/header') ?>
+
+<?php slot('toc') ?>
+<?php snippet(
+	'toc',
+	[
+		'title' => 'In this episode',
+		'items' => $page->layouts()->toBlocks()->filterBy('type', 'heading')
```
lukasbestle: `Issue`: The TOC links don't work for me. They point to the UUIDs of the blocks instead of their title slugs. We could either extend the TOC snippet to handle this case by generating the hash links for block objects or we could use `->map()` here to replace the IDs with the correct link targets.

lukasbestle (commented):
Works great for me except this one issue.

site/templates/kosmos-issue-virtual.php
texnixe: Thanks for spotting this, fixed with the last commit

lukasbestle (approved):

PR #2535, opened 2026-03-15
feat: Update Kosmos signup to use Mailcoach, update privacy page

texnixe:
## Description
This PR updates the current Kosmos subscription form to use Mailcoach instead of Newstroll. Also updates the privacy text.

Feel free to add additional changes as required.


lukasbestle:
A remaining use of Newstroll is in the `new-company` controller and its snippets. I am wondering if we should throw out either that complete page or at least the interactive elements. It's been four years since the company transition, don't think we need to keep updating it. However we would need to modify the hub to point users who have not yet agreed to email support then.

If we remove the logic on `new-company`, we can also throw out:

- `newstroll.list` key in `config.php`
- `site/plugins/newstroll`

texnixe:
If we throw out the complete page, the link I posted in the last Kosmos issue will go to the error page…, but we should probably throw out the interactive stuff and use the standard Kosmos form there as well.

texnixe:
Hm, looking at that page, it doesn't really make much sense anymore if we remove all the interactive stuff and then only keep a really bare-bones page. Maybe let's just remove the Kosmos subscription form for now (and the corresponding FAQ), then delete the complete page in 2 months or so (to prevent the error for newsletter readers).

texnixe:
How do we want to continue with this PR?

site/controllers/new-company.php
lukasbestle: This still uses the removed function `customerSubscribe`. The snippet `new-company/authenticated.php` can set the `newsletter` flag. People are still sent to the authenticated "new company" form when they log in to the hub and haven't migrated their data yet. So we either need to remove the newsletter checkbox and controller code for the authenticated form as well or we need to update it.

site/config/keys.sample.php
```diff
@@ -30,5 +29,12 @@
...
 			'clientSecret' => '',
 		]
-	]
+	],
+	'mailcoach'          => [
+		'emailList_uuid' => '',
+		'api_domain'     => '',
+		'api_key'        => '',
+		'api_url'        => '',
+		'form_url'       => '',
```
lukasbestle, line 38: `Suggestion`: We use camelCased option names everywhere else. For consistency I suggest we do the same here.

site/snippets/templates/kosmos/form.php
```diff
@@ -1,15 +1,29 @@
...
+		<input
+			type="hidden"
+			name="redirect_after_already_subscribed"
+			value="<?= 'https://' . option('keys.mailcoach.api_domain') . '/already-subscribed' ?>"
+		/>
+		<input
+			type="hidden"
+			name="redirect_after_subscription_pending"
+			value="<?= 'https://' . option('keys.mailcoach.api_domain') . '/redirect-after-pending' ?>"
+		/>
```
lukasbestle: `Question`: Do we need to override these redirect URLs? They look like the default URLs to me.

site/config/keys.sample.php
texnixe: These are now like in the other repos (partners, kosmos), so I'd prefer keeping them as they are?

site/snippets/templates/kosmos/form.php
```diff
@@ -1,15 +1,29 @@
...
 			<input class="input mb-3" id="email" name="email" required type="email">
-			<p class="text-sm">By subscribing you agree that we send you monthly newsletters. We won't ever spam you! You can unsubscribe at any time.</p>
+			<p class="text-sm">By subscribing you agree to receiving the Kosmos newsletter (usually once per month). We won't ever spam you! You can unsubscribe at any time.</p>
 		</div>
 		<div class="kosmos-field">
-			<label class="h5 mb-3 block" for="name">Name</label>
-			<input class="input mb-3" id="name" name="name" type="text">
-			<p class="text-sm">We use the German company Newstroll to collect subscribers and send out newsletters. Find out more in our <a class="underline" href="<?= url('privacy#newsletter') ?>">privacy policy</a>.</p>
+			<label class="h5 mb-3 block" for="first_name">First name</label>
+			<input class="input mb-3" id="name" name="first_name" type="text">
```
bastianallgeier, line 10:
```suggestion
			<input class="input mb-3" id="first_name" name="first_name" type="text">
```

content/privacy/text.txt
```diff
@@ -434,21 +434,19 @@ Fastmail is GDPR compliant. You can find the privacy policy of Fastmail here: <h
 
 ## Newsletter
 
-You can sign up to our monthly Kirby Kosmos newsletter to receive updates about Kirby, plugins, themes and other news around Kirby and the web. We never sign you up automatically. You have to use the sign up form on our website or during the Paddle checkout and confirm your subscription with the double-opt in process. To sign up we require and verify your email address and you can provide an optional name.
+You can sign up to our monthly Kirby Kosmos newsletter to receive updates about Kirby, plugins, themes and other news around Kirby and the web. We never sign you up automatically. You can sign up via the signup form on our website, or by activating the newletter checkbox during the Paddle checkout. You then need to confirm your subscription with the double-opt in process. To sign up, we require and verify your email address and you can provide an optional name.
```
bastianallgeier, line 437:
```suggestion
You can sign up to our monthly Kirby Kosmos newsletter to receive updates about Kirby, plugins, themes and other news around Kirby and the web. We never sign you up automatically. You can sign up via the signup form on our website, or by activating the newsletter checkbox during the Paddle checkout. You then need to confirm your subscription with the double-opt in process. To sign up, we require and verify your email address and you can provide an optional name.
```

content/privacy/text.txt
```diff
@@ -434,21 +434,19 @@ Fastmail is GDPR compliant. You can find the privacy policy of Fastmail here: <h
...
+2018 Antwerp
+Belgium*
 
-Newstroll offers analytics tools to check if newsletters have been opened and links have been clicked. We have disabled tracking on the subscriber level and only get access to the analytics for the entire list of subscribers. We only use those analytics to check how interesting and relevant our newsletters are over time. We never use them to analyze individual subscribers.
 
-You can unsubscribe from our newsletter at any time. This can be done by using the unsubscribe link in every newsletter or by sending an email to <support@getkirby.com> and we will handle it for you.
+You can unsubscribe from our newsletter at any time using the unsubscribe link in every newsletter, or by sending an email to <support@getkirby.com> and we will handle it for you.
 
-Newstroll complies with GDPR. You can find our more about their privacy policy on their site (German): https://www.newstroll.de/datenschutz/
+Mailcoach is fully compliant with GDPR. You can find our more about their privacy policy on their site: https://www.mailcoach.app/privacy-policy/
```
bastianallgeier, line 449:
```suggestion
Mailcoach is fully compliant with GDPR. You can find out more about their privacy policy on their site: https://www.mailcoach.app/privacy-policy/
```

site/snippets/templates/kosmos/form.php
bastianallgeier: `Nitpick`: We use HTML instead of XHTML style everywhere else. 

```suggestion
		<input
			type="hidden"
			name="redirect_after_subscribed"
			value="<?= 'https://' . option('keys.mailcoach.api_domain') . '/subscribed' ?>"
		>
		<input
			type="hidden"
			name="redirect_after_already_subscribed"
			value="<?= 'https://' . option('keys.mailcoach.api_domain') . '/already-subscribed' ?>"
		>
		<input
			type="hidden"
			name="redirect_after_subscription_pending"
			value="<?= 'https://' . option('keys.mailcoach.api_domain') . '/redirect-after-pending' ?>"
		>
```

bastianallgeier (commented):
I only found minor adjustments besides Lukas' notes.

site/snippets/templates/kosmos/form.php
texnixe: These have been removed

bastianallgeier (approved):

PR #2537, opened 2026-03-19
feat: Directly send signups to partners endpoint instead of Airtable

texnixe:
## Description
This PR is intended as the "partner" PR to https://github.com/getkirby/partners/pull/20, and sends data to the partner hub instead of to Airtable.

I also added a few more validation rules, but we might have to extend those even more for all fields.


site/config/keys.sample.php
```diff
@@ -30,5 +30,7 @@
 			'clientId'     => '',
 			'clientSecret' => '',
 		]
-	]
+	],
+	'signupToken' => '...',
```
bastianallgeier: I think we should use the same name in all places where we use those.

site/controllers/partners-signup.php
```diff
@@ -73,13 +72,26 @@
...
 			$page->validateReferences($plan, $references);
 			$page->validateWebsite($website);
 			$page->validateEmail($email);
 			$page->validateBusinessType($businessType);
+			$page->validateProjects($projects, $plan);
+			$page->validateDownloadLink($downloadLink);
 
 			// submit form values to Airtable
-			$response = Remote::post('https://api.airtable.com/v0/appeeHREbUMMaZGRP/tblrKOCF0cGAZmUQR', [
+			$response = Remote::post('http://partners.test/signup', [
```
bastianallgeier: The URL should be configurable.

bastianallgeier:
As a general thought, we have a couple rate limiting checks for the hub, which would also be helpful here to avoid bot attacks. @lukasbestle what do you think? We could maybe extract the code for those into a simple little internal plugin and use it here too.

site/controllers/partners-signup.php
texnixe: Oops, that was a leftover... we definitely don't want to post to .test

site/config/config.php
```diff
@@ -10,7 +10,8 @@
 	'hub'        => ['url' => 'https://hub.getkirby.com'],
 	'partners' => [
 		'url'        => 'https://partners.getkirby.com/partners.json',
-		'partnerUrl' => 'https://partners.getkirby.com/partners/'
+		'partnerUrl' => 'https://partners.getkirby.com/partners/',
+		'signupUrl'  => 'http://partners.getkirby.com/signup'
```
lukasbestle:
```suggestion
		'signupUrl'  => 'https://partners.getkirby.com/signup'
```

site/controllers/partners-signup.php
```diff
@@ -73,13 +72,26 @@
...
+
+			$page->validatePlan($plan);
 			$page->validateReferences($plan, $references);
 			$page->validateWebsite($website);
 			$page->validateEmail($email);
 			$page->validateBusinessType($businessType);
+			$page->validateProjects($projects, $plan);
+			$page->validateDownloadLink($downloadLink);
 
 			// submit form values to Airtable
```
lukasbestle:
```suggestion
			// submit form values to the partner hub
```

site/snippets/templates/partners-signup/signup.php
```diff
@@ -89,6 +89,11 @@
 [v-cloak] {
 	display: none !important;
 }
+
+.offset {
+	position:absolute;
+	left:-9999px;
```
lukasbestle: `Issue`: These will not hide the field and label from the accessibility tree. According to https://css-tricks.com/building-a-honeypot-field-that-works/#aa-preventing-bots-from-detecting-honeypots, `display: none` should do the trick.

site/snippets/templates/partners-signup/signup.php
```diff
@@ -100,7 +105,11 @@ class="columns dialog-form"
 		@submit="submit"
 	>
 		<input type="hidden" name="timestamp" :value="locale.timestamp">
-
+		<div class="offset">
+			<label for="date_of_birth">Date of birth</label>
+			<input type="text" name="date_of_birth" autocomple="off" class="offset" tabindex="-1">
```
lukasbestle:
```suggestion
			<input type="text" name="date_of_birth" autocomplete="off" class="offset" tabindex="-1">
```

But then again, these properties are not needed if we use `display: none`:

https://css-tricks.com/building-a-honeypot-field-that-works/#aa-unnecessary-enhancements

lukasbestle:
I agree on the rate limiting. The honeypot and extra validation might already help against bots, but in the past we had attacks that seemed human-controlled (because our time check was explicitly circumvented). Only a rate limit can definitely protect against that.

We could add that in a second step though IMO because nothing really changes here just for the switch from Airtable to partner hub.

Might make sense to add the rate limiting to the partner hub actually. This would then mean that we'd have to pass the client IP with the signup request, but the advantage would be that the rate limiting cache would not be on the site server.

site/snippets/templates/partners-signup/signup.php
texnixe: Hm, ok, my research said not to use `display: none`, that's why I implemented it like this which also what the Uniform honeypot field does.

bastianallgeier: I would assume too that display: none is quite easy to detect for a bot. I hate it when different articles contradict each other on the same topic. This is the honeypot field that is used by DreamForm, which seems to be working really well: 

```html
<input type="text" style="position:absolute;width:1px;height:1px;padding:0;margin:-1px;overflow:hidden;clip:rect(0,0,0,0);white-space:nowrap;border-width:0;" aria-hidden="true" tabindex="-1" name="website">
```

The interesting part here is that it uses inline styles. Could be to increase specificity? Not sure.

site/controllers/partners-signup.php
```diff
@@ -73,43 +71,56 @@
 			// there will be a redirect to Paddle)
 			Time::validate($timestamp);
 
+			if (get('date_of_birth')) {
+				http_response_code(200);
+				exit('OK');
+			}
+
+			if ((!$token = get('csrf')) || csrf($token) === false) {
```
lukasbestle: I think the CSRF check will not work in production because of the cache (both Kirby cache and KeyCDN cache).

If I understand our KeyCDN config correctly, we have set it so that cookies in the request are ignored and the page is rendered from cache anyway.

If we need the CSRF check, we would therefore need to set `header('Cache-Control: no-store');` for the page response so that KeyCDN never caches it and always retrieves it from the server. This however circumvents the purpose of the CDN.

Not sure how to proceed here.

site/snippets/templates/partners-signup/preview.php
```diff
@@ -214,14 +214,9 @@ class="field partner-listing-summary"
 	>
 		<label>
```
lukasbestle: `Question`: I saw you removed the field, but kept the label, which is now without field. Was this intended?

texnixe: Yes, that was intended, the short description doesn't make sense in the sign-up process. But probably shouldn't be a label but just text.

site/controllers/partners-signup.php
texnixe: Remove the token check for now

lukasbestle (approved):

PR #2542, opened 2026-06-14
Feature/prefill signup data after error

texnixe:
## Description
This PR makes makes sure to preserve user input data in case of form errors, so that users don't have to re-enter valid information. Invalid fields are cleared, though.

The validation method have been moved into a parent method, and now return a boolean. The purpose is to return multiple error messages at the top of the page if needed.

### DIFF
--- a/site/controllers/partners-signup.php
+++ b/site/controllers/partners-signup.php
@@ -24,7 +24,7 @@ return function (App $kirby, Page $page) {
 			'location'   => get('location'),
 			'website'    => get('website'),
 			'address'    => get('address'),
-			'projects'   => get('projects'),
+			'projects'   => (int) get('projects'),
 			'references' => get('references'),
 			'reviewRef'  => get('downloadLink'),
 			'contact'    => get('name'),
@@ -89,7 +89,7 @@ return function (App $kirby, Page $page) {
 				}
 
 				throw new Exception(
-					"This form contains the following errors:\n" .implode("\n", $messages)
+					"This form contains the following errors:\n" . 	implode("\n" , $messages)
 				);
 			}
 
@@ -98,16 +98,16 @@ return function (App $kirby, Page $page) {
 			// submit form values to the partner hub
 			$response = Remote::post(option('partners.signupUrl'), [
 				'data' => json_encode([
-					  'fields' => [
-						  'partnerstatus' => 'open',
-						  'people'        => $people,
-						  'price'         => $visitor->currencySign() . $localizedPrice,
-						  'checkout'      => $product->checkout('buy', $checkoutData),
-						  'created'       => date('Y-m-d H:i:s'),
-						  'ip'            => $kirby->visitor()->ip(hash: true),
-						  ...$data
-					  ]
-				  ], JSON_THROW_ON_ERROR),
+					'fields' => [
+						'partnerstatus' => 'open',
+						'people'        => $people,
+						'price'         => $visitor->currencySign() . $localizedPrice,
+						'checkout'      => $product->checkout('buy', $checkoutData),
+						'created'       => date('Y-m-d H:i:s'),
+						'ip'            => $kirby->visitor()->ip(hash: true),
+						...$data
+					]
+				], JSON_THROW_ON_ERROR),
 				'headers' => [
 					'Authorization' => 'Bearer ' . option('keys.partners.signupToken'),
 					'Content-Type'  => 'application/json',
@@ -157,8 +157,10 @@ return function (App $kirby, Page $page) {
 	}
 
 	// prefill form for renewals
-	if (($renew = param('renew')) &&
-		($partner = page('partners')->find($renew))) {
+	if (
+		($renew = param('renew')) &&
+		($partner = page('partners')->find($renew))
+	) {
 			$plan   = $partner->plan()->value();
 			$people = $partner->people()->value();
 	}
--- a/site/controllers/partners-signup.php
+++ b/site/controllers/partners-signup.php
@@ -18,18 +18,20 @@ return function (App $kirby, Page $page) {
 		$peopleNum    = max(1, min(4, (int)$people));
 		$plan         = get('plan');
 		$renew        = get('renew');
-		$businessName = get('businessName');
-		$businessType = get('businessType');
-		$location     = get('location');
-		$website      = get('website');
-		$address      = get('address');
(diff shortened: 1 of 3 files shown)


site/controllers/partners-signup.php
```diff
@@ -16,24 +16,26 @@
...
-		$discord      = get('discord');
-		$notes        = get('notes');
+		$data         = [
+			'plan'       => get('plan') ?? 'certified',
+			'title'      => get('businessName'),
+			'business'   => get('businessType'),
+			'location'   => get('location'),
+			'website'    => get('website'),
+			'address'    => get('address'),
+			'projects'   => get('projects'),
```
bastianallgeier: `Question` There was an (int) typecast on projects before. Is it no longer needed?

site/controllers/partners-signup.php
```diff
@@ -77,39 +79,35 @@
...
-			$page->validateProjects($projects, $plan);
-			$page->validateDownloadLink($downloadLink);
+			$errors   = $page->validate($data);
+			$messages = $page->getMessages();
+
+			if (count($errors)) {
+				foreach ($errors as $key => $value) {
+					// Clear form data for invalid fields
+					unset($data[$key]);
+				}
```
bastianallgeier: In site/models/partners-signup.php, the validate() method uses `businessType` and `downloadLink` in the returned errors array:

```php
$errors = [
    'businessType' => $this->validateBusinessType($data['business']),
    'downloadLink' => $this->validateDownloadLink($data['reviewRef']),
    ...
];
```

But above in the controller, the `$data` array uses different keys (business, reviewRef). 

```php
foreach ($errors as $key => $value) {
    unset($data[$key]); // tries to unset 'businessType' and 'downloadLink'
}
```

This looks wrong to me.

site/snippets/templates/partners-signup/signup.php
```diff
@@ -100,15 +100,7 @@ class="columns dialog-form"
...
-		<div style="position:absolute;width:1px;height:1px;padding:0;margin:-1px;overflow:hidden;clip:rect(0,0,0,0);white-space:nowrap;border-width:0;">
-			<label for="date_of_birth">Date of birth</label>
-			<input
-				type="text"
-				aria-hidden="true"
-				tabindex="-1"
-				name="date_of_birth"
-			>
-		</div>
+
```
bastianallgeier, line 112: `Question` is it correct that we no longer need the honeypot?

lukasbestle: I think we could keep the basic "date of birth" honeypot as first-level spam protection. We do have the rate limiting now, but without any honeypot, we could still get some spam submissions. I think the "date of birth" field doesn't have any disadvantages.

The timestamp honeypot is however no longer needed or actually harmful to actual users who need to resubmit the form. Sonja and I discussed this in Leipzig. I created a separate PR for this: https://github.com/getkirby/getkirby.com/pull/2545

site/controllers/partners-signup.php
```diff
@@ -77,39 +79,35 @@
...
-						'address'       => $address,
-						'projects'      => $projects,
-						'references'    => $references,
-						'reviewRef'     => $downloadLink,
-						'notes'         => $notes,
-						'created'       => date('Y-m-d H:i:s'),
-						'ip'            => $kirby->visitor()->ip(hash: true),
-					]
-				], JSON_THROW_ON_ERROR),
+					  'fields' => [
```
lukasbestle: `Minor issue`: This file has mixed indentation, could you please convert to tabs only?

site/controllers/partners-signup.php
```diff
@@ -77,39 +79,35 @@
...
+			$messages = $page->getMessages();
+
+			if (count($errors)) {
+				foreach ($errors as $key => $value) {
+					// Clear form data for invalid fields
+					unset($data[$key]);
+				}
+
+				throw new Exception(
+					"This form contains the following errors:\n" .implode("\n", $messages)
```
lukasbestle:
```suggestion
					"This form contains the following errors:\n" . implode("\n", $messages)
```

site/controllers/partners-signup.php
```diff
@@ -159,12 +157,12 @@
...
 
 	// prefill form for renewals
-	if ($renew = param('renew')) {
-		if ($partner = page('partners')->find($renew)) {
+	if (($renew = param('renew')) &&
+		($partner = page('partners')->find($renew))) {
 			$plan   = $partner->plan()->value();
 			$people = $partner->people()->value();
-		}
 	}
```
lukasbestle:
```suggestion
	if (
		($renew = param('renew')) &&
		($partner = page('partners')->find($renew))
	) {
		$plan   = $partner->plan()->value();
		$people = $partner->people()->value();
	}
```

site/models/partners-signup.php
```diff
@@ -6,83 +6,135 @@
...
 	{
-		$links = preg_split('/[\s,]+/', $references, -1, PREG_SPLIT_NO_EMPTY);
+		$links        = array_unique(preg_split('/[\s,]+/', $references, -1, PREG_SPLIT_NO_EMPTY));
 		$projectCount = $plan === 'regular' ? 2 : 4;
 
+		$links = array_unique($links);
+
 		if (count($links) < $projectCount) {
-			throw new Exception('A minimum number of ' . $projectCount . ' references is required');
+			$this->messages[] = 'A minimum number of ' . $projectCount . ' unique references is required';
```
lukasbestle: `Suggestion`: To avoid the messages in global state, the individual validation methods could return `string|bool` (with a string being an error message and `true` meaning "no error"). Then the `validate()` method could directly include the messages in its return array without the controller having to retrieve the messages separately.

lukasbestle (dismissed):

site/snippets/templates/partners-signup/signup.php
```diff
@@ -296,27 +297,27 @@ class="btn btn--filled"
 	form: {
 		// plan
 		people: "<?= $people ?? '1' ?>",
-		plan: "<?= $plan ?>",
+		plan: "<?= $data['plan'] ?? 'certified' ?>",
```
lukasbestle: `Question`: Does this not need to get escaped as well? It also comes from the request. It is validated, but the raw `$data` is passed to the template and therefore snippet even if the validation failed (to populate the form).

lukasbestle (approved):

PR #2544, opened 2026-06-16
feat: Do not serve dotfiles in nginx subdirectories

backflip:
## Description

This PR updates the nginx recipe to redirect not only dotfiles in the root, but in subdirectories, too.

### Summary of changes

Update of nginx example config.

### Reasoning

I stumbled upon this when moving the `accounts` folder outside of `site`: As it was no longer caught by `^/(content|site|kirby)/(.*)$`, I was able to access the `.htpasswd` files.


bastianallgeier (approved):

bastianallgeier:
Thanks for the fix!!

PR #2552, opened 2026-06-27
fix: PHP errors in site code

lukasbestle:
### Summary of changes

- Fix various issues that came up in the PHP error log in production.
- Clean up templates and pages that show up in `sitemap.xml`

### Reasoning

While reviewing the PHP error log I noticed that these issues kept coming up again and again. This was both a nuisance for users who then ran into errors on the site and also spammed our error log.

### Additional context

Best reviewed via individual commits because the changes don't interact with each other.


distantnative (approved):
I didn't test it directly, but read through each commit one by one. They all look good. I think this is fine to be merged.

PR #2559, opened 2026-07-18
Fixes to security advisory fetching via GitHub API

lukasbestle:
### Summary of changes

- Request all security advisories from GitHub API, not just the most recent 30
- Sort advisories with pending CVE ID to the top

### Reasoning

- We have reached the default limit of 30 results per API request. This limit can be increased with the `per_page` query param, but the proper solution is to follow the pagination links.
- With GitHub taking their time to assign CVE IDs due to the AI wave, it now happens that we need to release with pending CVE ID. Previously the `null` values were sorted to the bottom of the list, which meant that the most recent advisories are not as visible as they should be.


bastianallgeier (approved):
Works like a charm.

PR #2561, opened 2026-07-22
docs: Fix and complete permission keys

afbora:
The user blueprint reference had `access.settings` in an example. That key was renamed to `system` in 3.6 and the fallback was dropped in 3.8, so it has done nothing for years — but people kept copying it from the docs. This fixes that and fills in a few gaps around it.

Docs follow-up to getkirby/kirby#8319.

## Changes

**User blueprint reference** (`/docs/reference/panel/blueprints/user#permissions`)

- Fixed the `access` example: `settings: false` → `system: false`
- Added a note that `access.settings` was renamed in 3.6 and no longer works
- Added a short section about custom Panel areas (`access.<areaId>`)

**Permissions extension** (`/docs/reference/plugins/extensions/permissions`)

- Added a note that Panel areas get their permission automatically — you don't need this extension for them
- Explained where the category name comes from (`my/plugin` → `my.plugin`)
- Added an "Undefined permissions" section, since the deprecation warning links here

## Missing keys

I know these are outside the scope of getkirby/kirby#8319. While going through the page I noticed a few permissions were never documented:

- `files.read`
- `pages.read`
- `site.preview`
- the whole `languages` category (`create`, `delete`, `update`)

### DIFF
--- a/content/docs/2_reference/7_plugins/1_extensions/0_permissions/reference-extension.txt
+++ b/content/docs/2_reference/7_plugins/1_extensions/0_permissions/reference-extension.txt
@@ -6,6 +6,26 @@ Text:
 
 If you want to make the permissions for different actions of your plugin configurable by user role, you can register the available actions with the `permissions` extension.
 
+## Custom Panel areas
+
+If your plugin registers a Panel area, it automatically gets an `access.<areaId>` permission. You do not need the `permissions` extension for that.
+
+```php "/site/plugins/todos/index.php"
+Kirby::plugin('my/todos', [
+    'areas' => [
+        'todos' => fn () => [ /* ... */ ]
+    ]
+]);
+```
+
+```yaml "/site/blueprints/users/editor.yml"
+permissions:
+  access:
+    todos: false
+```
+
+Use the `permissions` extension only for actions inside your plugin (create, delete, publish …), not for access to the area itself.
+
 ## PHP definition
 
 Each available action needs to be set with the default permission value. The default value is used if the user's role does not specify the permissions for that action.
@@ -27,6 +47,8 @@ All permissions of the Kirby core are set to `true` by default. Only use the def
 ## Usage in the user blueprint
 
 The permissions can now be set in the user blueprint for each role (just like the (link: docs/guide/users/permissions text: core permissions)).
+
+Your permissions are grouped under a category named after the plugin, with the slash replaced by a dot: `my/permissionPlugin` becomes `my.permissionPlugin`. Use exactly that key in the user blueprint.
 
 ```yaml "/site/blueprints/users/editor.yml"
 title: Editor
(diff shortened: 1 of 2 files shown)


content/docs/2_reference/7_plugins/1_extensions/0_permissions/reference-extension.txt
```diff
@@ -6,6 +6,26 @@ Text:
 
 If you want to make the permissions for different actions of your plugin configurable by user role, you can register the available actions with the `permissions` extension.
 
+## Custom Panel areas
+
+If your plugin registers a Panel area, it automatically gets an `access.<areaId>` permission. You do **not** need the `permissions` extension for that.
```
distantnative:
```suggestion
If your plugin registers a Panel area, it automatically gets an `access.<areaId>` permission. You do not need the `permissions` extension for that.
```

content/docs/2_reference/7_plugins/1_extensions/0_permissions/reference-extension.txt
```diff
@@ -6,6 +6,26 @@ Text:
...
+]);
+```
+
+```yaml "/site/blueprints/users/editor.yml"
+permissions:
+  access:
+    todos: false
+```
+
+Use the `permissions` extension only for actions *inside* your plugin (create, delete, publish …), not for access to the area itself.
```
distantnative:
```suggestion
Use the `permissions` extension only for actions inside your plugin (create, delete, publish …), not for access to the area itself.
```

content/docs/2_reference/7_plugins/1_extensions/0_permissions/reference-extension.txt
```diff
@@ -28,6 +48,8 @@ All permissions of the Kirby core are set to `true` by default. Only use the def
 
 The permissions can now be set in the user blueprint for each role (just like the (link: docs/guide/users/permissions text: core permissions)).
 
+Your permissions are grouped under a category named after the plugin, with the slash replaced by a dot — `my/permissionPlugin` becomes `my.permissionPlugin`. Use exactly that key in the user blueprint.
```
distantnative:
```suggestion
Your permissions are grouped under a category named after the plugin, with the slash replaced by a dot: `my/permissionPlugin` becomes `my.permissionPlugin`. Use exactly that key in the user blueprint.
```

content/docs/2_reference/7_plugins/1_extensions/0_permissions/reference-extension.txt
```diff
@@ -43,4 +65,10 @@ You can then check the permissions for each action in your plugin:
...
 $kirby->user()->role()->permissions()->for('my.permissionPlugin', 'create') // false
-```
\ No newline at end of file
+```
+
+## Undefined permissions
+
+A permission that is neither a core permission nor registered through this extension is *undefined*. Kirby still applies undefined permissions for backwards compatibility, but they are not guaranteed to keep working — register them here instead.
+
+If a warning sent you here about an `access.*` key, you most likely want a core `access` key or a custom Panel area instead. See the (link: docs/reference/panel/blueprints/user#permissions text: user blueprint reference) for the full list.
```
distantnative: I would remove this section. I find it more confusing and the changes above make it very straightforward already in my opinion.

